PR #287, opened 2023-04-04
devcontainer: use official docker image

RobertGalatNordic:
use Nordic docker image from playground as base
KRKNWK-16674


ktaborowski (approved):
nice

totyz (approved):

PR #289, opened 2023-04-07
github: add comment with change in memory usage

RobertGalatNordic:
Every PR will now have automatic comment that shows the memory usage difference caused by the change, and absolute value of the memory used.

### DIFF
--- a/scripts/ci/combine_twister_reports.py
+++ b/scripts/ci/combine_twister_reports.py
@@ -1,0 +1,25 @@
+# Copyright (c) 2022 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+
+import json
+import sys
+
+if (len(sys.argv) <= 3):
+    print("not enough arguments")
+    exit(0)
+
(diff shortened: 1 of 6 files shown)


scripts/ci/combine_twister_reports.py
```diff
@@ -0,0 +1,25 @@
...
+
+combined = {"testsuites": []}
+
+for file in files_to_combine:
+    try:
+        with open(file, "r") as f:
+            report = json.load(f)
+            combined["testsuites"].extend(report.get("testsuites", []))
+    except:
+        print(f"Failed to open file {file}")
```
totyz, line 22:
```suggestion
for file in files_to_combine:
    try:
        with open(file, "r") as f:
            report = json.load(f)
            combined["testsuites"].extend(report.get("testsuites", []))
    except FileNotFoundError:
        print(f"Failed to open file {file}")
```

totyz (approved):
1 small improvement proposed

scripts/ci/combine_twister_reports.py
RobertGalatNordic: suggested change has been applied

PR #290, opened 2023-04-11
Make SX126x gpio assignments customizable

wabryan:
Adding the ability to change the SX126X radio's GPIO assignments for compatibility with other boards.

### DIFF
--- a/subsys/config/common/src/app_subGHz_config.c
+++ b/subsys/config/common/src/app_subGHz_config.c
@@ -98,16 +98,16 @@ const radio_sx126x_device_config_t radio_sx1262_cfg = {
 	.rx_boost = false,
 	.lna_gain = RADIO_RX_LNA_GAIN,
 	.bus_factory = &radio_spi_factory,
-	.gpio_power = 3,                // sx1262_NRESET
-	.gpio_int1 = 38,                // sx1262_DIO1
-	.gpio_radio_busy = 36,          // sx1262_BUSY
-	.gpio_rf_sw_ena = 42,           // sx1262 ANT_SW
+	.gpio_power = RADIO_SX126X_NRESET_GPIO,                // sx1262_NRESET
+	.gpio_int1 = RADIO_SX126X_DIO1_GPIO,                   // sx1262_DIO1
+	.gpio_radio_busy = RADIO_SX126X_BUSY_GPIO,             // sx1262_BUSY
+	.gpio_rf_sw_ena = RADIO_SX126X_ANT_SW_GPIO,            // sx1262 ANT_SW
 	.gpio_tx_bypass = 128,
 
 	.pa_cfg_callback = radio_sx1262_pa_cfg,
 
 	.bus_selector = {
-		.client_selector = 40, // sx1262_NSS
+		.client_selector = RADIO_SX126X_NSS_GPIO,      // sx1262_NSS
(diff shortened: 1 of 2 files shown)


RobertGalatNordic:
IMHO, that is good direction, but maybe this could be pushed a little further.

I would suggest creating configurations in https://github.com/nrfconnect/sdk-sidewalk/blob/main/Kconfig.dependecies with default values, and with no prompt, just like https://github.com/nrfconnect/sdk-sidewalk/blob/6815ffce7d898e8b1dcfe6fd5d22dc068451392a/Kconfig.dependecies#L179-L187, but it would be created if subghz configuration is enabled. This way the configuration would be present only if it is required, and because it would be hidden, it would not be possible to change with menuconfig (only with Kconfig.defaults of the applicaiton) This would also allow, to have multiple applications, one for DK, and one for a custom board with different pin configurations

wabryan:
So, again, maybe not the best implementation, but on our firmware side, I just added the following to `boards/arm/our_board/board.cmake` :

```
+add_compile_definitions(RADIO_SX126X_NRESET_GPIO=7)
+add_compile_definitions(RADIO_SX126X_DIO1_GPIO=27)
+add_compile_definitions(RADIO_SX126X_BUSY_GPIO=6)
+add_compile_definitions(RADIO_SX126X_ANT_SW_GPIO=8)
+add_compile_definitions(RADIO_SX126X_NSS_GPIO=5)
```

wabryan:
> This would also allow, to have multiple applications, one for DK, and one for a custom board with different pin configurations

I believe this solution satisfies these requirements. I attempted to make my changes as inconsequential as possible.

RobertGalatNordic:
Well  this solution may be convenient, but it would separate the configuration, and place some of it in Zephyr repository. I do not think that we will suggest modifying `boards/arm/our_board/board.cmake`. I think that the best place for this type of configuration would be in DeviceTree or in Kconfig. The Kconfig approach is definitely simpler to implement than DeviceTree, and it more closely resembles Your approach.
At the end of the day, we agree, that replacing those GPIO numbers with macros is the best approach. We will try to find the best solution to provide the convenient way for setting up those values.

wabryan:
Yeah, I hear you. I am not sure how to accomplish what you proposed without making a bunch of changes to the Zephyr and/or nRF repositories. Definitely open to any suggestions or ideas y'all have that make this as simple as possible.

wabryan:
Oh, I also misunderstood you. `boards/arm/our_board/board.cmake` lives within our repository, so there is no need to push anything to the Zephyr repository. I assume this would be true for other people too, but I guess I don't know that for sure.

We just add `list(APPEND BOARD_ROOT ${CMAKE_CURRENT_SOURCE_DIR})` to the top of the project's `CMakeLists.txt`

RobertGalatNordic:
Please take a look at https://github.com/nrfconnect/sdk-sidewalk/pull/294, We will go with gpio configuration via DTS

RobertGalatNordic:
https://github.com/nrfconnect/sdk-sidewalk/pull/294 resolved this issue, Please define DTS node with GPIOs required for the Semtech connection for your board. You can base the configuration on overlay files from `boards` directories in samples

wabryan:
https://github.com/nrfconnect/sdk-sidewalk/pull/294 seems to work, thank you!

PR #297, opened 2023-04-19
github: add badge with build status for main

RobertGalatNordic:
Badge has been added that show if all samples and tests build correctly on the latest commit on main branch


ktaborowski (approved):
nice thing, lgtm

PR #299, opened 2023-04-19
github: simplify memory usage table

RobertGalatNordic:
add option to print only elements that were different


ktaborowski (approved):
looks good

PR #307, opened 2023-04-27
DNM manifest: update nrf to upmerge pull request

ktaborowski:
[KRKNWK-16741]
https://github.com/nrfconnect/sdk-nrf/pull/10876


RobertGalatNordic:
samples do not build due to compilation warning `template_subghz/src/main.c:109:6: error: return type of 'main' is not 'int'`

PR #313, opened 2023-05-09
doc: update windows setup commands

RobertGalatNordic:
Update python command to install mfg tools


ktaborowski (approved):
In aws `python` command is used, byt `py` will work as well

PR #314, opened 2023-05-09
GitHub: add build artifacts to release

RobertGalatNordic:
attach buiild artifacts to relase

### DIFF
--- a/.github/workflows/tag_action.yml
+++ b/.github/workflows/tag_action.yml
@@ -8,15 +8,41 @@ on:
 jobs:
   build:
     permissions: write-all
-    name: Add documentaiton as asset
+    name: Create release page
     runs-on: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v3
-      - name: Build
-        run: cd doc; ./build_local.sh && zip -r documentation.zip build/html/*
 
-      - name: Upload binaries to release
+      - name: Check artifacts from commit
+        id: get_artifacts
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          commit: ${{github.sha}}
+          workflow: on_commit.yml
+          dry_run: true
+          if_no_artifact_found: ignore
+      
+      - name: Sleep 15 min to wait for builds
+        if: ${{steps.get_artifacts.outputs.dry_run == false}}
+        run: |
+          sleep 15m
+      
+      - name: Get artifacts from commit
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          commit: ${{github.sha}}
+          workflow: on_commit.yml
+      
+      - name: Build documentation
+        run: cd doc; ./build_local.sh && zip -r documentation.zip build/html/*
+  
+      - name: Pack build artifacts
+        run: |
+          zip -r build_artifacts.zip sample*
+          zip -r test_artifacts.zip tests*
+
+      - name: Upload documentation binaries to release
         uses: svenstaro/upload-release-action@v2
         with:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
@@ -24,4 +50,21 @@ jobs:
           asset_name: documentation.zip
           tag: ${{ github.ref }}
           overwrite: true
-          body: "HTML documentation is includded in the release asset in documentation.zip file."
+      
+      - name: Upload build binaries to release
+        uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: build_artifacts.zip
+          asset_name: twister_build_artifacts.zip
+          tag: ${{ github.ref }}
+          overwrite: true
+
+      - name: Upload test binaries to release
+        uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: test_artifacts.zip
+          asset_name: twister_test_artifacts.zip
+          tag: ${{ github.ref }}
+          overwrite: true


.github/workflows/tag_action.yml
```diff
@@ -3,25 +3,68 @@ name: Publish release
...
+    name: Create release page
     runs-on: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v3
-      - name: Build
+
+      - name: Check artifacts from commit
+        id: get_artifacts
+        uses: dawidd6/action-download-artifact@v2
```
ktaborowski, line 19: [dawidd6](https://github.com/dawidd6) ?

.github/workflows/tag_action.yml
```diff
@@ -3,25 +3,68 @@ name: Publish release
...
+        with:
+          commit: ${{github.sha}}
+          workflow: on_commit.yml
+          dry_run: true
+          if_no_artifact_found: ignore
+      
+      - name: Sleep 15 min to wait for builds
```
ktaborowski, line 26: Shouldn't we start that action in flow after the build instead of waiting fixed amount of time?

.github/workflows/tag_action.yml
```diff
@@ -3,25 +3,68 @@ name: Publish release
...
+        if: ${{steps.get_artifacts.outputs.dry_run == false}}
+        run: |
+          sleep 15m
+      
+      - name: Get artifacts from commit
+        uses: dawidd6/action-download-artifact@v2
```
ktaborowski, line 32: [dawidd6](https://github.com/dawidd6) ?

.github/workflows/tag_action.yml
```diff
@@ -3,25 +3,68 @@ name: Publish release
...
+          commit: ${{github.sha}}
+          workflow: on_commit.yml
+      
+      - name: Build documentation
         run: cd doc; ./build_local.sh && zip -r documentation.zip build/html/*
+  
+      - name: Pack build artifacts
+        run: |
+          zip -r build_artifacts.zip sample*
+          zip -r test_artifacts.zip tests*
```
ktaborowski, line 43: Good idea :)

.github/workflows/tag_action.yml
```diff
@@ -3,25 +3,68 @@ name: Publish release
...
         with:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
           file: doc/documentation.zip
           asset_name: documentation.zip
           tag: ${{ github.ref }}
           overwrite: true
-          body: "HTML documentation is includded in the release asset in documentation.zip file."
+      
+      - name: Upload build binaries to release
+        uses: svenstaro/upload-release-action@v2
```
ktaborowski, line 55: https://github.com/svenstaro ?

.github/workflows/tag_action.yml
```diff
@@ -3,25 +3,68 @@ name: Publish release
...
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: build_artifacts.zip
+          asset_name: twister_build_artifacts.zip
+          tag: ${{ github.ref }}
+          overwrite: true
+
+      - name: Upload test binaries to release
+        uses: svenstaro/upload-release-action@v2
```
ktaborowski, line 64: https://github.com/svenstaro ?

.github/workflows/tag_action.yml
```diff
@@ -3,25 +3,68 @@ name: Publish release
 on:
   push:
     tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+'
+      - 'v[0-9]+.[0-9]+.[0-9]+**'
```
ktaborowski, line 6: What is `**` for?

ktaborowski (commented):
Its nice to have a better release page
Some questions below:

.github/workflows/tag_action.yml
RobertGalatNordic: This step is executed only if the build did not finish yet. If the artifacts were found in previous step, this sleep is omited.

It would be nice to trigger this action automatically is that the tag can be created at any time.
So we can not trigger this action after the commit build has finished (because we do not know if the tag has been created on this commit) 

So the current trigger is the most accurate - This action will be executed when the tag is pushed to repo, but if we do merge, and quickly push the tag, the on commit plan will not finish yet, and the artifacts will be missing. In that case waiting 15min, is no such long time, and it ensures that the build finishes ( it takes <10 min)

.github/workflows/tag_action.yml
RobertGalatNordic: the same as with the previous comment, some actions are not official, and most of them can be done by GitHub API, but why reinvent the wheel :/

.github/workflows/tag_action.yml
RobertGalatNordic: This will allow for longer versioning schemes so now it will allow for example "v1.2.3.4" the restricion is that it has to start with "v", and have 3 numbers separated by dot, next there can be anything

.github/workflows/tag_action.yml
ktaborowski: https://github.com/nrfconnect/sdk-nrf/blob/e7b53c3a17b61339fb144d5c9f8290bcd49c0ac0/.github/workflows/docpublish.yml#L15 I see nrf also uses that. so it's ok for me

.github/workflows/tag_action.yml
ktaborowski: Based on [nrf history](https://github.com/nrfconnect/sdk-nrf/tags) I would rather expect -rcX, -devX instead of next number. 

```suggestion
      - 'v[0-9]+.[0-9]+.[0-9]+(-[a-z]+[0-9]+)*'
```

.github/workflows/tag_action.yml
ktaborowski: sound reasonable
maybe there is a better solution, but lets give this one a try - if it fails we will investigate for the better one

ktaborowski (approved):

PR #318, opened 2023-05-12
GitHub: increase paralel matrix for slowest job

RobertGalatNordic:
Building step 2/3 of tests takes almost twice the time of other steps. It is therefore the bottle neck, by spliting this job to more tasks, it will distribute the load more evently and reduce the build time


ktaborowski (approved):
lgtm

PR #320, opened 2023-05-17
Scripts: fix generation of hardware map for twister UT

totyz:
recover boards for both modes AUTO and user_devconf passed as parameter to run_ut.sh 
remove additional debug port for the newest nrf52 boards 
do not use nrf52840 <2.0.0
use logging module

Signed-off-by: Tomasz Tyzenhauz <tomasz.tyzenhauz@nordicsemi.no>

### DIFF
--- a/scripts/ci/fill_hardware_map.py
+++ b/scripts/ci/fill_hardware_map.py
@@ -7,7 +7,13 @@ import subprocess
 
 import yaml
 import argparse
+import logging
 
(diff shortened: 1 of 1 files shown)


totyz:
UT on HW passed on: node 60, 70 (problematic), 58,94,128

RobertGalatNordic (approved):
lgtm

PR #343, opened 2023-05-29
vscode: add lldb and doxygen to devenvironment

RobertGalatNordic:
lldb enalbes debugging for native posix.
If clangd is used, plugin C/C++ which usualy provides debugger for native application is disabled. LLDB is used instead in that case.

Doxygen helps with commenting functions and file headers.


ktaborowski (approved):
lgtm

PR #358, opened 2023-06-05
doc: added compatibility matrix

annwoj:
related to KRKNWK16963

### DIFF
--- a/doc/compatibility_matrix.rst
+++ b/doc/compatibility_matrix.rst
@@ -1,0 +1,15 @@
+.. _compatibility_matrix:
+
+Compatibility matrix
+====================
+
+In this section, you can check the SDK version dependencies for the nRF52840 SoC.
+The following table demonstrates the compatibility between the nRF Connect SDK and Sidewalk MCU SDK versions:
+
++-----------------------------------------+-------------------------+-------------------------------------------------+-----------------------------+
+| Nordic Semiconductor's Sidewalk release | nRF Connect SDK version | Amazon specification (Sidewalk MCU SDK version) | Amazon qualification status |
++=========================================+=========================+=================================================+=============================+
+| v1.14.3                                 | v2.3.0                  | v1.14                                           | Qualified/approved          |
++-----------------------------------------+-------------------------+-------------------------------------------------+-----------------------------+
+| v1.14.4                                 | v2.4.0                  | v1.14                                           | Planned                     |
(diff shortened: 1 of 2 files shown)


annwoj:
@ktaborowski @wojciechpluszynski I moved the table to a separate section on the main page, however, I'm thinking that maybe it could as well go under the Software requirements (?) (https://nrfconnect.github.io/sdk-sidewalk/setting_up_sidewalk_environment/requirements.html#software-requirements) . Please let me know whichever you prefer. 
Also, I modified headings a bit and added some introduction lines - let me know if anything needs to be corrected.

doc/index.rst
```diff
@@ -13,6 +13,7 @@ You can browse the content by referring to the following pages:
    :caption: Subpages:
 
    about_sidewalk.rst
+   compatibility_matrix.rst
```
ktaborowski, line 16: consider to move it down, near release notes

annwoj: I guess that depends on when would a user need this information. If it's not required at the very beginning of setting everything up it can be moved down. Is that the case?

doc/compatibility_matrix.rst
```diff
@@ -0,0 +1,15 @@
+.. _compatibility_matrix:
+
+Compatibility matrix
+====================
+
+Before you start working with the Sidewalk environment, familiarize yourself with the outlined version dependencies.
+The following table demonstrates the compatibility between the versions of the nRF Connect SDK and Sidewalk MCU SDK:
+
```
ktaborowski, line 8: minor: maybe add an information it's for nrf52840 chip?

ktaborowski (approved):
looks good, as for me it is better to have separated menu item for compatibility matrix

doc/index.rst
ktaborowski: Hmm, I though at the first time developers needs to set up, build and test the sdk. And when is works, they looks on compatibility matrix and release notes when it goes to final product. This matrix is not for getting started, but for the productization

ktaborowski: But is is also ok for me to left it as is

annwoj: Makes sense - thanks for the explanation :) will move it to right above the release notes section

doc/compatibility_matrix.rst
annwoj: added to the line above

PR #359, opened 2023-06-05
crypto: optimize crypto algorithms

RobertGalatNordic:
KRKNWK-17003
Signed-off-by: Robert Gałat <robert.galat@nordicsemi.no>


Kconfig.dependencies
```diff
@@ -50,11 +50,16 @@ config SIDEWALK_CRYPTO
 	imply PSA_WANT_ALG_CTR
 	imply PSA_WANT_ALG_GCM
 	imply PSA_WANT_ALG_CMAC
+	imply PSA_WANT_ALG_ECB_NO_PADDING
+	imply PSA_WANT_ALG_SHA_1
 	imply PSA_WANT_ALG_SHA_256
+	imply PSA_WANT_ALG_SHA_512
```
MarGasiorek, line 56: The SHA-512 is not used directly ([labcollab](https://issues.labcollab.net/browse/HALO-24093?focusedCommentId=27623181&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27623181)). I think that in this case this config is not necessary anymore, and in code you can use this snippet: 
#ifdef  CONFIG_PSA_WANT_ALG_SHA_512
https://github.com/nrfconnect/sdk-sidewalk/blob/8b27fcbc5a40b5ed4a26410e867e849652d6751d/subsys/sal/sid_pal/src/sid_crypto.c#L424C2-L426
#endif /* CONFIG_PSA_WANT_ALG_SHA_512 */

Kconfig.dependencies
```diff
@@ -50,11 +50,16 @@ config SIDEWALK_CRYPTO
 	imply PSA_WANT_ALG_CTR
 	imply PSA_WANT_ALG_GCM
 	imply PSA_WANT_ALG_CMAC
+	imply PSA_WANT_ALG_ECB_NO_PADDING
+	imply PSA_WANT_ALG_SHA_1
```
MarGasiorek, line 54: Did you try build without this config? I'm not sure is it necessary, especially when this algorithm is not safe and [NIST](https://csrc.nist.gov/news/2022/nist-transitioning-away-from-sha-1-for-all-apps) recommends migration to another standards.

PR #363, opened 2023-06-09
ACK and DUP example

wabryan:


wabryan:
I know this isn't the best way to do this, but I am not sure what the appropriate route is since this repository doesn't have the ability to open "Issues." So, apologies there. At any rate, in this small modification to the `template_subghz` sample, I am looking into how exactly the rx and tx attributes work, and what is expected behavior versus actual behavior.

Running the sample, and then pressing the "Hello" button once leads to:

```

[00:00:12.309,814] <inf> application_state: time_sync = true
[00:00:12.309,967] <inf> callbacks: Device Is registered, Time Sync Success, Link status: {BLE: Down, FSK: Down, LoRa: Up}
[00:00:12.310,058] <inf> callbacks: Link mode on lora = {Cloud: True, Mobile: False}
[00:00:20.942,932] <inf> button: button pressed 3 short
[00:00:20.943,450] <inf> application_state: sending = true
[00:00:20.943,450] <inf> board_events: queued data message id:3
[00:00:23.008,850] <inf> application_state: sending = false
[00:00:23.008,880] <inf> callbacks: sent message(type: 2, id: 3)
[00:00:23.008,880] <inf> callbacks: sent message attributes: rx_ack_request:0 rx_is_msg_ack:0 tx_request_ack:0
[00:00:27.289,062] <inf> application_state: receiving = true
[00:00:27.289,123] <inf> application_state: receiving = false
[00:00:27.289,123] <inf> callbacks: received message(type: 3, link_mode: 1, id: 3 size 0 is_msg_ack: 1 is_dup: 0)
[00:00:27.289,154] <inf> callbacks: Message data: [00:00:32.271,728] <inf> application_state: receiving = true
[00:00:32.271,789] <inf> application_state: receiving = false
[00:00:32.271,820] <inf> callbacks: received message(type: 3, link_mode: 1, id: 3 size 0 is_msg_ack: 1 is_dup: 0)
[00:00:32.271,850] <inf> callbacks: Message data: 

```

I am specifically curious about the `msg_desc_attr.rx_attr.ack_requested`, `msg_desc_attr.rx_attr.is_msg_ack`, and  `msg_desc_attr.tx_attr.request_ack` fields. 

Is it expected behavior to have the `msg_desc_attr.tx_attr.request_ack` attribute be wiped out after the message is sent? I.e. the callback function will have no way of knowing it sent a message that requested an ACK.

Is it expected behavior to have the `msg_desc_attr.rx_attr.is_msg_duplicate` field not show that a duplicate ACK message was received?

Let me know if there's a better place for these kinds of questions. This seems too niche of a question for the DevZone.

Thank you.

RobertGalatNordic:
GitHub's issues and discussions are disabled, because we prefer to use DevZone for all questions and requests, so do not hesitate to post even the more technical questions there. The PR are left open for the contributions to the project, so if You want to introduce some improvement or fix a bug, feel free to create PR.

Regarding Your question, callback functions are implementations of Sidewalk API, so the call arguments are controlled by Sidewalk library.

I find a little odd that You try to pick the value of `rx_attr` and `tx_attr`. Those fields are part of a union type `msg_desc_attr`, so they reference the same memory region.

https://docs.sidewalk.amazon/sid-api-dg/ The Sidewalk developer guide may help You understand how the Sidewalk API is designed.

Looking on the 
https://github.com/nrfconnect/sdk-sidewalk/blob/8b27fcbc5a40b5ed4a26410e867e849652d6751d/subsys/sal/common/sid_ifc/sid_api.h#L230-L235
You can see that those fields are only valid for two cases:
You fill `tx_attr` when You intend to pass the message descriptor to `sid_put_msg`

and when You receive the callback `on_msg_received` You can look in `rx_attr`

In case of callback `on_msg_sent` it seems like the descriptor is used only to reference the message type and ID, that will allow identifying the message that has been sent, other fields seems to be empty, and we can see this in logs that You provided.

Be aware that receiving `on_msg_sent` callback does not guarantee that the message got through to the cloud. It only signals the application, that the message left the device, and was transmitted to the gateway.

ktaborowski:
closing this issue - if you have more question please ask via devzone forum

wabryan:
Okay, so generally working as designed for the ACK attributes. Still wondering about the `is_msg_duplicate` field, though.

PR #367, opened 2023-06-14
scripts: Extend run_ut script by selecting platform to use

totyz:
Add new parameter -p which can be used to narrow twister to work only with selected platforms. It can be passed multiple times.


RobertGalatNordic (approved):
lgtm

PR #369, opened 2023-06-15
test: use k_sleep for delay in timer test

RobertGalatNordic:
KRKNWK-17144
replaced waiting loop based on counter for platform agnostic k_sleep


ktaborowski (approved):
lgtm

PR #395, opened 2023-09-01
tests: implement functional test in twister

RobertGalatNordic:
Add simple test to sid_dut application
to cycle the initialization and deinitializaiton procedure

This test is in prototype state, it has been created to check the functionality of twister funcitonal tests.


samples/sid_dut/pytest/test_functional.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+load_dotenv(dotenv_path=Path(__file__).parent.joinpath('.env.secrets'))
+load_dotenv(dotenv_path=Path(__file__).parent.joinpath('.env.constants'))
+
+MFG_HEX_PATH = os.getenv('MFG_HEX_PATH')
+
+FAST_RESPONSE_TIMEOUT = int(os.getenv('FAST_RESPONSE_TIMEOUT'))
+FLASH_MFG_TIMEOUT = int(os.getenv('FLASH_MFG_TIMEOUT'))
+DEFAULT_TIMEOUT = int(os.getenv('DEFAULT_TIMEOUT'))
```
ktaborowski, line 27: Why to get those values as env variables? Couldn't it be defined here? Or just keept in one config file (.json or .yml)

samples/sid_dut/pytest/test_functional.py
```diff
@@ -0,0 +1,158 @@
...
+        BLE,
+        # FSK,
+        # LORA,
+    ],
+)
+def test_functional_init_start_stop_deinit_multiple(dut: DeviceAbstract, transport):
+    logger.debug('start of bug_test')
+
+    log_output(dut)
+    time.sleep(1)
```
ktaborowski, line 117: Why to sleep if next step is wait for prompt?

samples/sid_dut/pytest/test_functional.py
```diff
@@ -0,0 +1,158 @@
...
+def log_output(dut: DeviceAbstract):
+    for line in dut.iter_stdout:
+        logger.debug("#: " + line)
+        if line == "":
+            return
+
+
+BLE = 1
+FSK = 2
+LORA = 3
```
ktaborowski, line 102: I'm not convinced to define transports in such a way. Maybe a simple dictionary of transports will be better? or a simple class?

ktaborowski (changes requested):
Interesting concept, definitely worth further exploration
For sure a documentation is missing - at lease a simple readme file which tells how to prepare and run this test.
I think some parts are overengineered for such simple test, my proposition is to create a simple test to start and stop the sidewalk transport and another one which loop the first test. Also you can remove flashing a mfg hex for now - this can be a precondition in readme like echo and semteach shield presence, and this test will be much cleaner (we can add mfg flash later, as more functional test will be added this way).

RobertGalatNordic:
Closing this pull request. This was created to evaluate the use of PyTest integration with twister as a tool for debugging and development. This solution proved useful for reproducing bugs, and eventually fixing them.

The PR is a little rough, and in deed it does not meet our quality standards. Integration of PyTests run by twister will need a lot more work to be ready for release, so this PR is to be closed, but it will remain in GH history as reference to what can be done. 
We will continue to evaluate the use of PyTest for fast bug reproduction and local testing, as an alternative for functional tests run by CI.

PR #412, opened 2023-09-27
nRF53 reduce current consumption

ktaborowski:
Reduce power consumption in template_ble for nRF53
And some minor improvements in samples configurations


samples/template_ble/src/main.c
```diff
@@ -107,6 +114,14 @@ static void app_setup(void)
 		}
 	}
 #endif
+
+#if defined(CONFIG_NORDIC_QSPI_NOR)
+	const struct device *const qspi_dev = DEVICE_DT_GET(EXTERNAL_FLASH);
+
+	if (device_is_ready(qspi_dev)) {
+		pm_device_action_run(qspi_dev, PM_DEVICE_ACTION_SUSPEND);
```
RobertGalatNordic, line 122: Should we include the suspension of external flash in all our samples ? Maybe this could be tied to a DFU code as we need external flash only for dfu, and we need to wake if from suspension only when a new image is downloaded to the device? We could extract this to a common module and make it available for all samples that use DFU (and those who do not should have external flash disabled)?

samples/template_ble/boards/nrf5340dk_nrf5340_cpuapp.overlay
```diff
@@ -1,11 +1,28 @@
 /*
- * Copyright (c) 2022 Nordic Semiconductor ASA
+ * Copyright (c) 2024 Nordic Semiconductor ASA
```
RobertGalatNordic, line 2: Code from future, maybe we could make a template to autofill the header for new files, this would solve the issue of incorrect year in the copyright

RobertGalatNordic (approved):

samples/template_ble/src/main.c
ktaborowski: Good question, I think it should stay here (at least for now) for two reasons:

1. The code is related by device name with the dts overlay file in sample folder. So it's not just code to move (the customer may want to change the flash device it is should be easy to change in code).
2. I think this should be a part of zephyr power management (maybe a kconfig for nordic_qspi_nor?). It will be more valuable to have it common to all ncs samples, not just sidewalk.

PR #421, opened 2023-10-03
GHA: fix issue with reporting fail build as pass

RobertGalatNordic:
In Build step after the actual build preformed by twister, there were commands that will always return status 0, therefore this step always returns pass


ktaborowski (approved):
looks good

PR #431, opened 2023-10-11
sample: added log to sidewalk message callbacks

RobertGalatNordic:
KRKNWK-17750
in our tests we should wait for this callback to confirm delivery This log message also shows all the parameters of the message


ktaborowski (approved):
Removed a lot of duplicated code, good job :)
Consider if additional debugs are worth dut sample size increase.

PR #435, opened 2023-10-17
tests: nRF52  wont suspend flash to save power

ktaborowski:
In power tests external flash will be suspend only on nRF53. This allows to perform usb dfu on nRF52 in Sidewalk mode


totyz (approved):
lgtm

PR #441, opened 2023-10-26
Release doc

ktaborowski:


annwoj (approved):
I had something different in mind for release notes and known issues so we can exclude those edits for now - tomorrow I will create a separate PR for that (wanted to wait for the ncs doc merge), other than this LTGM :)

PR #455, opened 2023-11-13
gpio: refactor gpio to support new board

RobertGalatNordic:
make gpio module platform independent, and enable transfering only selected gpios to the Sidewalk API


subsys/config/common/src/app_subGHz_config.c
```diff
@@ -135,27 +136,36 @@ static radio_sx126x_device_config_t radio_sx1262_cfg = {
...
-	radio_sx1262_cfg.gpio_tx_bypass = sid_gpio_utils_get_gpio_number_dt(
+	radio_sx1262_cfg.gpio_tx_bypass = sid_gpio_utils_register_gpio(
 			(struct gpio_dt_spec)GPIO_DT_SPEC_GET_OR(DT_NODELABEL(semtech_sx1262_tx_bypass), gpios, INVALID_DT_GPIO));
-
+	
+	sid_pal_gpio_set_direction(radio_sx1262_cfg.gpio_radio_busy, SID_PAL_GPIO_DIRECTION_INPUT);
+	sid_pal_gpio_set_direction(radio_sx1262_cfg.gpio_int1, SID_PAL_GPIO_DIRECTION_INPUT);
+	sid_pal_gpio_set_direction(radio_sx1262_cfg.bus_selector.client_selector, SID_PAL_GPIO_DIRECTION_OUTPUT);
+	sid_pal_gpio_set_direction(radio_sx1262_cfg.gpio_power, SID_PAL_GPIO_DIRECTION_OUTPUT);
+	sid_pal_gpio_set_direction(radio_sx1262_cfg.gpio_rf_sw_ena, SID_PAL_GPIO_DIRECTION_OUTPUT);
```
ktaborowski, line 167: as we discussed offline - this is a side effect in get config function. please move to pin set config / initialization function

also `#include "sid_pal_gpio_ifc.h"` can be removed then

ktaborowski (changes requested):

subsys/config/common/src/app_subGHz_config.c
RobertGalatNordic: removed as discussed

ktaborowski (approved):
looks good to me, please make sure there is no regression in subghz samples

PR #461, opened 2023-11-20
tests: use valid gpio for nss

RobertGalatNordic:
the use of sid_gpio api helps with test portability by making it independent of physical HW configuration


ktaborowski (approved):
Good change

PR #491, opened 2024-02-14
BLE: add option to print RSSI on received message

RobertGalatNordic:
KRKNWK-18493
The RSSI it tied to the connection, and it is taken from soft device. Disabling the averaging of the RSSI may increase power consumption see sdc_hci_cmd_vs_set_power_control_request_params() for side effects.

### DIFF
--- a/subsys/sal/sid_pal/src/sid_ble_ama_service.c
+++ b/subsys/sal/sid_pal/src/sid_ble_ama_service.c
@@ -7,6 +7,10 @@
 /** @file sid_ble_ama_service.c
  *  @brief Bluetooth low energy Amazon service implementation.
  */
+#include <zephyr/kernel.h>
+
+#include <zephyr/bluetooth/hci.h>
+#include <sdc_hci_vs.h>
 
 #include <sid_ble_ama_service.h>
 #include <sid_ble_adapter_callbacks.h>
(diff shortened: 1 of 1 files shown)


subsys/sal/sid_pal/src/sid_ble_ama_service.c
```diff
@@ -47,7 +53,25 @@ static ssize_t ama_srv_on_write(struct bt_conn *conn, const struct bt_gatt_attr
 	ARG_UNUSED(conn);
 	ARG_UNUSED(offset);
 	ARG_UNUSED(flags);
+#if CONFIG_SIDEWALK_BLE_RSSI_READ_ON_MESSAGE
+#if CONFIG_SIDEWALK_BLE_RSSI_AVERAGE_DISABLE
+	// avg[n] = gamma * avg[n - 1] + (1 - gamma) * rssi[n]
+	// Here, gamma equals beta/4096, and rssi[n] equals the current RSSI.
+	// leave the last rssi in the average.
```
ktaborowski, line 60: how is this comment related with code?

subsys/sal/sid_pal/src/sid_ble_ama_service.c
```diff
@@ -47,7 +53,25 @@ static ssize_t ama_srv_on_write(struct bt_conn *conn, const struct bt_gatt_attr
...
+	const sdc_hci_cmd_vs_set_power_control_request_params_t parameters_for_rssi_calculation = {
+		.beta = 4096
+	};
+	sdc_hci_cmd_vs_set_power_control_request_params(&parameters_for_rssi_calculation);
+#endif // CONFIG_SIDEWALK_BLE_RSSI_AVERAGE_DISABLE
+
+	uint16_t conn_handle = 0;
+	bt_hci_get_conn_handle(conn, &conn_handle);
+	const sdc_hci_cmd_vs_read_average_rssi_t p_params = { .conn_handle = conn_handle };
```
ktaborowski, line 69: why is it `const`?

Kconfig.dependencies
```diff
@@ -279,4 +279,15 @@ config SIDEWALK_DFU_SERVICE_BLE_DEFAULTS
 config MCUMGR_TRANSPORT_BT_AUTHEN
 	default n
 
+config SIDEWALK_BLE_RSSI_READ_ON_MESSAGE
+	bool "Read RSSI on received message"
+	help
+	  Add log message with the current average value of RSSI on every received BLE message
+
+config SIDEWALK_BLE_RSSI_AVERAGE_DISABLE
```
ktaborowski, line 287: Isn't It sounds like something configurable on softdevice controller level. Isn't this config redundant to some NCS BLE SDC config?

RobertGalatNordic:
This PR was just an example how to get RSSI value for the connection in NCS. The Current API does not expect this value, so there is no need to keep this change in the source code.

PR #498, opened 2024-02-16
Fix crypto heap size in samples

ktaborowski:
Increase mbedtls heap size like in tests

```
[00:02:00.926,879] <inf> sid_ble_conn: BT Connected
[00:02:02.303,802] <err> sid_crypto: PSA Error code: -141
[00:02:02.305,084] <err> sid_ble: Advertising failed to start (err -12)
[00:02:02.346,282] <inf> sid_ble_conn: BT Disconnected Reason: 0x16 = LOCALHOST_TERM_CONN
```
fixes PSA error -141, ble advertising is still visible


samples/template/Kconfig.defconfig
```diff
@@ -60,8 +60,10 @@ config SHELL_STACK_SIZE
 	default 8192 if TEMPLATE_APP_CLI
 
 config HEAP_MEM_POOL_SIZE
-	default 8192 if SIDEWALK_ON_DEV_CERT
-	default 1024
+	default 4096
+
+config MBEDTLS_HEAP_SIZE
+	default 4096
```
RobertGalatNordic, line 66: I do not think we should have this in the template's Kconfig, 
IMO the better place would be the sidewalk/Kconfig.dependencies under the CONFIG_SIDEWALK_CRYPTO, because this change is needed always when we need to use CONFIG_SIDEWALK_CRYPTO, and not just in template

tests/functional/crypto/Kconfig
```diff
@@ -14,18 +14,22 @@ config SIDEWALK_CRYPTO_LOG_LEVEL
...
-	default 8192
+	default 4096
+
 config MBEDTLS_ENABLE_HEAP
 	default y
-config MBEDTLS_HEAP_SIZE
-	default 8192
 
+config MBEDTLS_HEAP_SIZE
+	default 4096
```
RobertGalatNordic, line 32: And here we would like to get rid of the explicitly setting up those values, but this is a part of the bigger test refactoring task, so it may be left as suggested for now, and addressed during test refactoring

RobertGalatNordic (commented):
Need to discuss where to place the default value of the heap size for crypto

samples/template/Kconfig.defconfig
ktaborowski: Agree, but when defined in Kconfig, the default value is covered at 512 from other kconfig

totyz (changes requested):
NOK.
Still problem is visible on nrf52 during registration

RobertGalatNordic (approved):

PR #501, opened 2024-02-19
semtech_cfg: Add trim cap callback

RobertGalatNordic:
enable setting value of the trim cap for semtech driver


ktaborowski (approved):
lgtm

PR #529, opened 2024-04-04
Krknwk 18773 dfu errors

ktaborowski:
Based on https://github.com/nrfconnect/sdk-sidewalk/pull/521
Created new PR to avoid breaking the base one.

Increased flash operation timeout, fix for error during erasing secondary firmware partition.


totyz (approved):
DFU manual tests PASSED. I don't see any problems here so far except long ~90s image swap time

PR #542, opened 2024-04-12
shell: replace log messages in on dev cert to shell print

RobertGalatNordic:
Because the shell is also responsible for handling log messages, it is impossible to do so while handling the shell command. This results in the delay between the creation of the log message inside the shell handler and its process and printing. Without this message the other side can not be sure if the command succeded, so it can not proced with next command wich causes significant delay.

Now the shell printer is used to pass the messages from shell command and it does no go through logging system.

## CI parameters

(branch, hash, pull/XXX/head)
NRF_revision=main

manifest-pr-skip (remove this line to create PR in NRF)

## Description

JIRA ticket: KRKNWK-18381

Detail description of the change

## Self review

- [ ] There is no TODO/FIXME comments without associated issue ticket.
- [ ] Commits are properly organized
- [ ] Verification
  - [ ] Unit tests were updated to include the change.
  - [ ] Change has been tested.

### DIFF
--- a/utils/src/sid_on_dev_cert_cli.c
+++ b/utils/src/sid_on_dev_cert_cli.c
@@ -78,7 +78,8 @@ static struct sid_base64_ctx cert_app_key_base64_ctx = {
 	.state = 0,
 };
 
-static sid_error_t sid_on_dev_cert_cli_print_base64(const uint8_t *input, size_t input_size)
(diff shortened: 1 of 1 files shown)


totyz (approved):
Works.
TAS updated

ktaborowski (approved):

PR #563, opened 2024-05-21
Upmerge 16.05.2024

RobertGalatNordic:
new SMF api has more arguments.

## CI parameters

```yaml
Github_actions:
  NRF_revision: pull/15351/head

  Create_NRF_PR: false
Jenkins:
  test-sdk-sidewalk: sdk-nrf-pr-15351
```

## Description

https://github.com/nrfconnect/sdk-nrf/pull/15351

## Self review

- [ ] Change has been tested.


samples/sid_end_device/child_image/mcuboot/prj.conf
ktaborowski, line 1: Just asking: is it intentional to remoce boot baner also form debug configurations?

samples/sid_end_device/sysbuild/hci_ipc/prj.conf
ktaborowski, line 1: minor: this is not dts for nrf54, but config for nrf53 netcore (misleading commit message)

ktaborowski (approved):
Loogs good overall

samples/sid_end_device/child_image/mcuboot/prj.conf
RobertGalatNordic: Yes, the boot banner increases the size of the bootloader, and therefore it does not fit in our dedicated space, so this also affects the debug versions of the application (I consider debug in terms of the debug version of the application, not the whole package, but in the future I would like to move away from custom, optimized size of the bootloader partition in favor of the default - configuration that could be common across different projects)

samples/sid_end_device/sysbuild/hci_ipc/prj.conf
RobertGalatNordic: yes this should be in previous commit, it had to slip when I added files to commits

PR #575, opened 2024-06-20
sample: Refactor event system for sampes

RobertGalatNordic:
## CI parameters

```yaml
Github_actions:
  NRF_revision: main

  Create_NRF_PR: false
Jenkins:
  test-sdk-sidewalk: master
```

## Description

JIRA ticket: KRKNWK-18976
Enum based events were replaced by functions.
This allows to simplify the event handling,
isolate events from each other in functions
and keep them separated from other code.

The enums for events were replaced by macros that define an event object. The objects are created in place, and are reasonably small. Because of that the user code does not require modification.

## Self review

- [ ] Change has been tested.


samples/sid_end_device/include/sidewalk.h
```diff
@@ -9,30 +9,79 @@
...
+};
 typedef struct {
 	struct sid_handle *handle;
 	struct sid_config config;
 	struct sid_status last_status;
 } sidewalk_ctx_t;
 
+typedef struct {
+	const char *name;
+	void (*call[STATE_LAST])(void *ctx, void *state_machine);
```
ktaborowski, line 28: this should have a typedef

samples/sid_end_device/include/sidewalk.h
```diff
@@ -9,30 +9,79 @@
...
+} sidewalk_event_t;
+typedef struct {
+	sidewalk_event_t event_handler;
+	void *ctx;
+} sidewalk_ctx_event_t;
+
+typedef struct sm_s {
+	struct smf_ctx ctx;
+	struct smf_state sid_states[STATE_LAST];
```
ktaborowski, line 37: Why is a whole table of states needed byt he state machine struct now?

samples/sid_end_device/CMakeLists.txt
```diff
@@ -15,6 +15,8 @@ target_include_directories(app PRIVATE include)
 target_sources(app PRIVATE
     src/main.c
     src/sidewalk.c
+    src/sidewalk_state_event_handlers.c
```
ktaborowski, line 18: minor: make names shorter

samples/sid_end_device/include/cli/app_shell_events.h
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#ifndef APP_DUT_H
+#define APP_DUT_H
+
+#include <sidewalk.h>
+
+void dut_sidewalk_event_init_handler(void *ctx, void *state_machine);
```
ktaborowski, line 12: minor: make names shorter

samples/sid_end_device/include/cli/app_shell_events.h
```diff
@@ -0,0 +1,55 @@
...
+void dut_sidewalk_event_get_mtu_handler(void *ctx, void *state_machine);
+void dut_sidewalk_event_get_time_handler(void *ctx, void *state_machine);
+void dut_sidewalk_event_get_status_handler(void *ctx, void *state_machine);
+void dut_sidewalk_event_get_option_handler(void *ctx, void *state_machine);
+void dut_sidewalk_event_set_option_handler(void *ctx, void *state_machine);
+void dut_sidewalk_event_set_dest_id_handler(void *ctx, void *state_machine);
+void dut_sidewalk_event_set_conn_req_handler(void *ctx, void *state_machine);
+
+#define DUT_EVENT_INIT                                                                             \
+	APP_SIDEWALK_EVENT(DUT_EVENT_INIT, { [STATE_SIDEWALK] = dut_sidewalk_event_init_handler })
```
ktaborowski, line 25: I don't understand , to avoid using enums, you create an event mechanism based on fuctions pointers, but you add another enums (defines). This seems to be more compincated, not simplified.

ktaborowski (changes requested):
Maybe I should better understand the details, just a biref overwiew now.
Not bad, jsut needs to be clarified before merge.

samples/sid_end_device/src/sidewalk.c
```diff
@@ -121,265 +108,40 @@ static void state_sidewalk_entry(void *o)
...
-	if (sm->event.id >= SID_EVENT_LAST) {
-		app_dut_event_process(sm->event, sm->sid);
+	if (!validate_event(sm, STATE_SIDEWALK)) {
+		return;
 	}
-#endif /* CONFIG_SID_END_DEVICE_CLI */
+	LOG_DBG("Handle event %s", sm->application_event.event_handler.name ?
+					   sm->application_event.event_handler.name :
+					   "UNKNOWN");
+	sm->application_event.event_handler.call[STATE_SIDEWALK](sm->application_event.ctx, sm);
```
ktaborowski, line 139: minor: consider simlification of this place. It is one of the most importatnt place

ktaborowski (commented):
Review summary, please consider to:
* Use shorter names
* Restructure structures, keep it simple
* Refactor also events in sensor monitoring tx thread

samples/sid_end_device/include/sidewalk.h
RobertGalatNordic: done

samples/sid_end_device/include/sidewalk.h
RobertGalatNordic: is is easier to access the states for transition. 
previously the event handlers were in the same file as the global state table, now they can be anywhere.
Because I pass the `sm_t` struct to the event handler, it is easy to access states for transition no matter where the actual hanlder is placed.

samples/sid_end_device/include/cli/app_shell_events.h
RobertGalatNordic: Renamed the handlers now they follow the schema `state` _ event _ `Event name`

samples/sid_end_device/include/cli/app_shell_events.h
RobertGalatNordic: This is a definition of an event. It can be emitted in any state, but different state will handle this event differently.
That is why Event defines the handle for every possible state (if it is needed, if the event must be ignored just ommit the handler for that state)

samples/sid_end_device/src/sidewalk.c
RobertGalatNordic: `call[STATE]` is an actual event handler, it is inside the event object retrieved from the queue next to the context to pass to this event, and this event from queue is stored in the state machine object. 

I tried to do some aliases to simplify this call chain

totyz:
Stress tests: https://jenkins-ncs.nordicsemi.no/job/KRKNG/job/Sidewalk/job/test-sdk-sidewalk/job/test-branch/128/

RobertGalatNordic:
TODO:
- [ ] rebase
- [ ] sid_dut events in DFU state do not handle heap data
- [ ] add ctx_clean function to event (and call it if present in fsm state handler after event handler) Event handler can ignore the source of ctx, and if it is needed the cleanup will free the context even if there is no handler
- [ ] consider creating an object for events, and use the event macros to return pointer - allow modification of the event (in case new state is created, and some event need to be handled in this new state? )

ktaborowski (changes requested):
After review I have some doubts regarding refactoring event system with the proposed macros: it increases flexability, but it's more error prone and harder to understand for newcomers.
Please cosider redesign the refactored structures and corespoding macros.

RobertGalatNordic:
new approach is explored in PR #578

PR #599, opened 2024-08-30
tests: Add Sidewalk crypto algorithm perforamnce tests.

MarGasiorek:
Sidewalk crypto algorithm perforamnce tests.

## CI parameters

```yaml
Github_actions:
  NRF_revision: main

  Create_NRF_PR: false
Jenkins:
  test-sdk-sidewalk: master
```

## Description

JIRA ticket: 

## Self review

- [ ] Change has been tested.

### DIFF
--- a/subsys/sal/sid_pal/src/sid_crypto.c
+++ b/subsys/sal/sid_pal/src/sid_crypto.c
@@ -22,6 +22,41 @@
 
 LOG_MODULE_REGISTER(sid_crypto, CONFIG_SIDEWALK_CRYPTO_LOG_LEVEL);
 
+#ifdef CONFIG_SIDEWALK_CRYPTO_PERF_TEST
+
+#define CRYPTO_PERF_START(fmt, ...)                                                                \
+	do {                                                                                       \
+		LOG_INF(fmt, ##__VA_ARGS__);                                                       \
+		step_cnt = 0;                                                                      \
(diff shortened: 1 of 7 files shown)


subsys/sal/sid_pal/src/sid_crypto.c
```diff
@@ -22,6 +22,41 @@
 
 LOG_MODULE_REGISTER(sid_crypto, CONFIG_SIDEWALK_CRYPTO_LOG_LEVEL);
 
+#ifdef CONFIG_SIDEWALK_CRYPTO_PERF_TEST
```
ktaborowski, line 25: I'm not feeling comforatable with adding a special code to this module just for purpose of the one test.
If we need to measure Sidewalk performance, we should opearte on Sidewalk (sid_pal) api in tests.
If we need to measure ncs crypto (psa api), we should create a test in ncs.
But it's just a comment to open discussion, not necessary a blocker for merge.

ktaborowski (commented):
I have a general problem with this test, is it realy a Sidewalk functional test?
I see a point with testing, and code itself looks good at the first look, just a question of it's place in repo

ktaborowski (changes requested):
Should not moficy sid_crypto.c only for testing purposes

ktaborowski:
Closing - might be continued in the future

PR #603, opened 2024-09-10
pal: switch to multi-advertising BLE API

konradderda:
This commit prepares PAL to be used alongside other BLE instances.

## CI parameters

```yaml
Github_actions:
  NRF_revision: main

  Create_NRF_PR: false
Jenkins:
  test-sdk-sidewalk: master
```

## Description

JIRA ticket: KRKNWK-19430

## Self review

- [X] There is no commented code.
- [X] There are no TODO/FIXME comments without associated issue ticket.
- [X] Commits are properly organized.
- [ ] Change has been tested.


konradderda:
Regarding `Change has been tested.` tick - The device has registered and sent test message to cloud but I do not possess enough experience to test the change thoroughly.

RobertGalatNordic:
good job, but Sidewalk has also API for disabling BLE, I think that we should have some kind of global reference counter and do not disable the BLE stack if some other part of the application uses it ( https://github.com/nrfconnect/sdk-sidewalk/blob/7238295f476f125127a9889e1981eccfd876b62d/subsys/sal/sid_pal/src/sid_ble_adapter.c#L254-L267
)

Kconfig
```diff
@@ -195,6 +195,14 @@ config SIDEWALK_BLE_ADV_INT_SLOW
 	range 20 10240
 	default 1000
 
+config SIDEWALK_BLE_ID
+	int "Set Sidewalk's BLE identity identifier"
+	range 1 255
```
MarGasiorek, line 200: I'm curious is it possible to set  `SIDEWALK_BLE_ID` to e.g 32 when `BT_ID_MAX` is set to 2?
May be here should be something like:
```suggestion
	range 1 BT_ID_MAX
```

konradderda: Due to the implementation of BT stack BT_ID_MAX has to be higher than a max number of needed IDs, e.q. to use 2 IDs we need to set BT_ID_MAX to at least 3.

I have added a BUILD_ASSERT to compare these values

RobertGalatNordic:
This implementation does not introduce paralel SMP and Sidewalk connection, but lays a ground for such implementaiton by separating and filtering advertising and connections based on ID.

The parallel Sidewalk and SMP service will be made by separate PR, as it needs more sophisticated way of calling bt_disable to make sure that sidewalk does not disable BLE when it switches transport  while SMP is working on DFU

RobertGalatNordic (approved):

ktaborowski:
Merged in https://github.com/nrfconnect/sdk-sidewalk/pull/643

PR #616, opened 2024-09-24
sample: set STATIC_INIT_GNU for the sidewalk sample

tejlmand:
The sidewalk sample linking places stack_protector from the C library in the init_array sections.

For proper initialization in Zephyr then content in this init_array section requires STATIC_INIT_GNU to be configured.

## CI parameters

```yaml
Github_actions:
  NRF_revision: pull/16860/head

  Create_NRF_PR: false
Jenkins:
  test-sdk-sidewalk: master
```

## Description

JIRA ticket: 

## Self review

- [X] There is no commented code.
- [X] There are no TODO/FIXME comments without associated issue ticket.
- [X] Commits are properly organized.
- [X] Change has been tested.

### DIFF
--- a/samples/sid_end_device/prj.conf
+++ b/samples/sid_end_device/prj.conf
@@ -3,6 +3,7 @@
 #
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
+CONFIG_STATIC_INIT_GNU=y
 
 # Sidewalk
 CONFIG_SIDEWALK=y
--- a/samples/sid_end_device/prj_release.conf
+++ b/samples/sid_end_device/prj_release.conf
@@ -3,6 +3,7 @@
 #
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
+CONFIG_STATIC_INIT_GNU=y
 
 # Sidewalk
 CONFIG_SIDEWALK=y
--- a/tests/manual/simple_bootloader/pm_static_nrf52840dk_nrf52840.yml
+++ b/tests/manual/simple_bootloader/pm_static_nrf52840dk_nrf52840.yml
@@ -1,8 +1,73 @@
+app:
+  address: 0x7200
+  end_address: 0xfd000
+  region: flash_primary
+  size: 0xf5e00
+external_flash:
+  address: 0xf6000
+  end_address: 0x800000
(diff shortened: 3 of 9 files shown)


samples/sid_end_device/prj.conf
```diff
@@ -3,6 +3,7 @@
 #
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
+CONFIG_STATIC_INIT_GNU=y
```
ktaborowski, line 6: build test failed (github actions):
samples/sid_end_device/prj.conf:6: warning: attempt to assign the value 'y' to the undefined symbol STATIC_INIT_GNU

ktaborowski (changes requested):

RobertGalatNordic:
@tejlmand I do not know if You are aware, but there is an issue with toolchain.
In the toolchain v2.7.0 (the latest avaliable toolchain) the python verison is `Python 3.9.18` 
But the change introduced by the upmerge in zephyr require python 3.10 >.

The exact change is here: 
zephyr/scripts/pylib/twister/twisterlib/statuses.py#L24-L37

```
    @staticmethod
    def get_color(status: TwisterStatus) -> str:
        match(status):
            case TwisterStatus.PASS:
                color = Fore.GREEN
            case TwisterStatus.SKIP | TwisterStatus.FILTER | TwisterStatus.BLOCK:
                color = Fore.YELLOW
            case TwisterStatus.FAIL | TwisterStatus.ERROR:
                color = Fore.RED
            case TwisterStatus.STARTED | TwisterStatus.NONE:
                color = Fore.MAGENTA
            case _:
                color = Fore.RESET
        return color
```

The notation `match(status):` is unsupported in python on version 3.9, and is available only from 3.10.

This line was recently changed by `Lukasz Mrugala <lukaszx.mrugala@intel.com>` .

So just to summarize, the zephyr version listed in https://github.com/nrfconnect/sdk-nrf/blob/939805062e20e9feaea7529ccb55158da63b4c2b/west.yml#L75
contain change in file zephyr/scripts/pylib/twister/twisterlib/statuses.py:26, that is incompatible with python 3.9, and that is the version of python shipped with the latest available toolchain

RobertGalatNordic (approved):

PR #626, opened 2024-10-17
Storage: use ZMS for new RRAM devices

RobertGalatNordic:
nrf54 has RRAM and therefore should use ZMS

## CI parameters

```yaml
Github_actions:
  NRF_revision: pull/17994/head

  Create_NRF_PR: false
Jenkins:
  test-sdk-sidewalk: master
```

## Description

JIRA ticket: 

## Self review

- [ ] Change has been tested.


rlubos (approved):

tomi-font (approved):

Kconfig.dependencies
```diff
@@ -138,7 +138,12 @@ config SIDEWALK_MFG_STORAGE_DIAGNOSTIC
 config SIDEWALK_STORAGE
 	bool
 	default SIDEWALK
-	imply NVS
+	imply NVS if !SOC_FLASH_NRF_RRAM
```
ktaborowski, line 141: minor: isn't there a config for the traditional flash? It would be more futureproof not tu use a negation in this case

CMakeLists.txt
```diff
@@ -9,3 +9,9 @@ if(CONFIG_SIDEWALK_BUILD)
 	add_subdirectory(subsys)
 	add_subdirectory(utils)
 endif()
+
+if(CONFIG_SOC_SERIES_NRF54LX AND CONFIG_SETTINGS_NVS)
+    message(WARNING "
+    The nRF54L series is not optimized for the NVS settings backend due to RRAM specifications.
+    Consider using the CONFIG_SETTINGS_ZMS Kconfig option, which is recommended for this type of non-volatile memory.")
```
ktaborowski, line 16: minor: could you make it shorter?

tests/unit_tests/pal_storage_kv/testcase.yaml
```diff
@@ -1,7 +1,19 @@
 tests:
-  sidewalk.unit_tests.storage_kv:
+  sidewalk.unit_tests.storage_kv.NVS:
     sysbuild: true
     platform_allow: native_posix
     tags: Sidewalk
     integration_platforms:
       - native_posix
+
+  sidewalk.unit_tests.storage_kv.ZMS:
```
ktaborowski, line 9: Do we need such testcase in mainline?

ktaborowski (approved):
Looks good, please have a look at questions below

tomi-font:
FYI, integrating this in https://github.com/nrfconnect/sdk-nrf/pull/17994.

tests/unit_tests/pal_storage_kv/testcase.yaml
RobertGalatNordic: Well now we will use two backends for settings, and alto this test runs on native_posix, I think that we should not exclude one of the used backends, simply because it is newer.

Maybe  we should consider if this test is useful at all. Perhaps we should drop it, and rely only on the one run on real hardware.

Kconfig.dependencies
RobertGalatNordic: Maybe, but this could cause an issue, if both of the configs are enabled, it would be hard to track which one would be selected as backend for Settings. This way, it is clear, that they are mutually exclusive

RobertGalatNordic: and matter uses the same scheme to select storage backend

totyz (approved):

PR #630, opened 2024-10-21
subGHz: fix SPI configuration

RobertGalatNordic:
KRKNWK-19506
add working configuration for nrf54l15dk

## CI parameters

```yaml
Github_actions:
  NRF_revision: main

  Create_NRF_PR: false
Jenkins:
  test-sdk-sidewalk: enable-nrf54l-subghz
```

## Description

JIRA ticket: 

## Self review

- [x] Change has been tested.


subsys/sal/sid_pal/src/sid_gpio_utils.c
```diff
@@ -78,6 +78,8 @@ uint32_t sid_gpio_utils_register_gpio(struct gpio_dt_spec gpio_from_dts)
 						      .pin = gpio_from_dts.pin },
 					    .configuration_cache = gpio_from_dts.dt_flags };
 
+	sid_gpio_utils_gpio_set_flags(ctx.next_free_slot,
```
ktaborowski, line 81: What are those flags for?

doc/setting_up_sidewalk_environment/setting_up_hardware.rst
```diff
@@ -101,67 +101,36 @@ Connect the Semtech shield to the Arduino-compatible headers on the development
...
+      | ANT_SW        | P2.10        |
+      +---------------+--------------+
+      | SPI MOSI      | P0.1         |
+      +---------------+--------------+
+      | SPI MISO      | P0.2         |
+      +---------------+--------------+
+      | SPI SCK       | P0.3         |
+      +---------------+--------------+
+      | NRESET        | P2.8         |
+      +---------------+--------------+
```
ktaborowski, line 133: minor: consider table rows permutation
```suggestion
      +---------------+--------------+
      | SX126X shield | nRF54L15 DK  |
      +===============+==============+
      | DIO1          | P1.11        |
      +---------------+--------------+
      | BUSY          | P2.6         |
      +---------------+--------------+
      | NRESET        | P2.8         |
      +---------------+--------------+
      | ANT_SW        | P2.10        |
      +---------------+--------------+
      | NSS           | P0.0         |
      +---------------+--------------+
      | SPI MOSI      | P0.1         |
      +---------------+--------------+
      | SPI MISO      | P0.2         |
      +---------------+--------------+
      | SPI SCK       | P0.3         |
      +---------------+--------------+

```

RobertGalatNordic: ok, good suggestion

subsys/sal/sid_pal/src/sid_gpio_utils.c
RobertGalatNordic: This is to set the default flags of the gpio when it is registered for a first time. It should be redundant as we pull those flags from DTS, but I call it just to be sure that the state of new gpio is defined.  (previously it was hidden behind a static guard, so essentially we set this only on first GPIO registerd, but not for any other).

So to summarize, The change is not big, it only expands the behaviour to all GPIOs regardles of the order of registering, and as far as I understand this operation should not be even necessary.

We can remove it in next code refactoring session (if it won't break anything and tests pass) ;)

totyz (approved):
Tested manually FSK on all supported SoCs. No problems found

ktaborowski (approved):

PR #643, opened 2024-11-13
Make smp and sidewalk parallel

RobertGalatNordic:
## CI parameters

```yaml
Github_actions:
  NRF_revision: main

  Create_NRF_PR: false
Jenkins:
  test-sdk-sidewalk: master
```

extends on #603
## Description

JIRA ticket: KRKNWK-17860, KRKNWK-18460

## Self review

- [ ] Change has been tested.


subsys/sal/common/sid_ifc/bt_app_callbacks.c
ktaborowski, line 1: I think this is not a good place dor this file (in this directory there are files to be changed when Sidewalk libraries are updates) - move  to `sid_pal/include`?

subsys/sal/common/sid_ifc/bt_app_callbacks.h
ktaborowski, line 1: I think this is not a good place dor this file (in this directory there are files to be changed when Sidewalk libraries are updates) - move  to `sid_pal/include`?

subsys/sal/common/sid_ifc/bt_app_callbacks.c
```diff
@@ -0,0 +1,43 @@
...
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+#include <bt_app_callbacks.h>
+#include <zephyr/bluetooth/bluetooth.h>
+#include <stdbool.h>
+#include <zephyr/kernel.h>
+
+static uint32_t bt_enable_count = 0;
+
+int app_bt_enable(bt_ready_cb_t cb)
```
ktaborowski, line 13: In genreal, I would recommend to keep the naming convention of Sidewalk module. The `app_` prefix is used for sample configuration like `app_ble_config`. For ble code we used `sid_ble_` prefixes (`sid_ble_advert`, `sid_ble_connectopn`)

subsys/sal/sid_pal/src/sid_ble_adapter.c
```diff
@@ -213,14 +214,35 @@ static sid_error_t ble_adapter_set_tx_pwr(int8_t tx_power)
...
+	if (count > BT_ID_SIDEWALK) {
+		return 0;
+	}
+
+	do {
+		ret = bt_id_create(NULL, NULL);
+		if (ret < 0) {
+			return ret;
+		}
+	} while (ret != BT_ID_SIDEWALK);
```
ktaborowski, line 233: risk of inifinite loop?

Why isn't the id saved in some static struct and compared in callback? I see something like this is done in nrodic_dfu and some sources in ncs.

subsys/sal/sid_pal/src/sid_ble_adapter.c
```diff
@@ -404,8 +438,10 @@ static sid_error_t ble_adapter_deinit(void)
 {
 	LOG_DBG("Sidewalk -> BLE");
 	sid_ble_conn_deinit();
-
-	int err = bt_disable();
+	sid_ble_advert_deinit();
```
ktaborowski, line 441: Why to add advertisiement deinit here?

subsys/sal/sid_pal/src/sid_ble_connection.c
```diff
@@ -33,6 +34,17 @@ static struct bt_conn_cb conn_callbacks = {
 
 static struct bt_gatt_cb gatt_callbacks = { .att_mtu_updated = ble_mtu_cb };
 
+static bool should_handle_event(struct bt_conn *conn)
```
ktaborowski, line 37: rename e.g. 'ble_connect_is_valid`

* use file specific prefix
* don't use should

+missing comment doc

Also it's incosinstent with the nordic_dfu module...

subsys/sal/common/sid_ifc/bt_app_callbacks.h
```diff
@@ -0,0 +1,50 @@
...
+ * @return int result from @bt_disable or 0 if app_bt_enable has been called more than app_bt_disable
+ */
+int app_bt_disable();
+
+
+/**
+ * @brief BT ids used for extended advertising. 
+ * This allows to identify connections from different extended adverticements.
+ */
+enum BT_id_values{
```
ktaborowski, line 39: don't use uppercase in type names

subsys/sal/common/sid_ifc/bt_app_callbacks.h
```diff
@@ -0,0 +1,50 @@
...
+        _BT_ID_DEFAULT = BT_ID_DEFAULT,
+        BT_ID_SIDEWALK,
```
ktaborowski, line 41:
```suggestion
        BT_ID_SIDEWALK = BT_ID_DEFAULT +1,
```

subsys/sal/common/sid_ifc/bt_app_callbacks.h
```diff
@@ -0,0 +1,50 @@
...
+        #if defined (CONFIG_SIDEWALK_DFU)
+        BT_ID_SMP_DFU,
+        #endif
+        _BT_ID_MAX
+};
+
+BUILD_ASSERT(_BT_ID_MAX <= CONFIG_BT_ID_MAX, "Too many BT Ids! increase CONFIG_BT_ID_MAX, to match _BT_ID_MAX");
```
ktaborowski, line 48: I like the idea to get the fast feedback to developer during build, but I think the BLE module should assert when you try to use too many ids.
My recommendation is to remove this, to keep code simple, and easier to maintain

subsys/sal/sid_pal/src/sid_ble_connection.c
```diff
@@ -41,12 +53,18 @@ static struct bt_gatt_cb gatt_callbacks = { .att_mtu_updated = ble_mtu_cb };
  */
 static void ble_connect_cb(struct bt_conn *conn, uint8_t err)
 {
+	const bt_addr_le_t *bt_addr_le;
```
ktaborowski, line 56: not initlized

utils/sidewalk_dfu/nordic_dfu.c
```diff
@@ -63,10 +100,52 @@ static void led_action(struct k_timer *timer_id)
...
 {
-	LOG_ERR("DFU did not start or could not complete. Reset to exit dfu mode");
-	sys_reboot(SYS_REBOOT_COLD);
+	LOG_ERR("DFU did not start or could not complete. Exit dfu mode");
+}
+
+static void connected(struct bt_conn *conn, uint8_t err)
+{
+	struct bt_conn_info cinfo;
+	int ec;
```
ktaborowski, line 109:
```suggestion
	int ret;
```

ktaborowski (changes requested):
Just a first-look review. Code looks good in gerenal, but naming and file location is incositent withe the rest of the repo. Plus some question about work.
To be continued.

subsys/sal/common/sid_ifc/bt_app_callbacks.h
RobertGalatNordic: I think if we removed this build assert, we would find out we do not have enough BT_ID_MAX in runtime, when we try to get new ID, and IMO that is a little too late.

subsys/sal/sid_pal/src/sid_ble_connection.c
RobertGalatNordic: Ok, I can address the styling issues,

The nordic_dfu module is a little easier, because there is easier access to advertising structure with the ID, where as in Sidewalk, advertising and connections are in separate files, so it is harder to access it by. But As I mentioned, ID is just an index, and it is application responsiblity to manage what service to connect with what ID, so I think it is acceptable that we use enum to filter the connection ( as we use the same enum to insert the id in advertising)

subsys/sal/sid_pal/src/sid_ble_adapter.c
RobertGalatNordic: Because we can free the advertising from BLE stack. Maybe for now it is not very useful, but I figured out if we asked for some resources, it would be nice to give them back when we are done working with them.

subsys/sal/common/sid_ifc/bt_app_callbacks.h
RobertGalatNordic: agree, this will be moved

utils/sidewalk_dfu/nordic_dfu.c
RobertGalatNordic: This was copied from mesh sample, as I hope it will solve one of the issues I see in Jenkins, if so, I will refactor this code

ktaborowski (approved):
Ok, no need to change form me

Kconfig.dependencies
```diff
@@ -38,6 +38,10 @@ config SIDEWALK_BLE
 	help
 		Sidewalk Bluetooth Low Energy (BLE) module
 
+config SIDEWALK_BLE_NAME
+	string "BLE name adverticed for Sidewalk"
+	default "SID_APP"
```
ktaborowski, line 44: I would recommend to left the name for Sidewalk advertising as it was, and rename dfu advertising

samples/sid_end_device/src/sensor_monitoring/app.c
```diff
@@ -230,6 +237,73 @@ void app_start_tasks(void)
...
+			return false;
+		}
+		if (bt_uuid_cmp(attr->uuid,
+				BT_UUID_DECLARE_128(LOG_EXAMPLE_CHARACTERISTIC_UUID_VAL_NOTIFY)) ==
+		    0) {
+			LOG_WRN("block LOG_EXAMPLE_CHARACTERISTIC_UUID_VAL_NOTIFY in DFU connection");
+			return false;
+		}
+	}
+#endif //defined(CONFIG_SIDEWALK_DFU)
```
ktaborowski, line 307: not small amount of repreated code, consider make it a function?

samples/sid_end_device/src/sensor_monitoring/app.c
```diff
@@ -264,6 +338,12 @@ void app_start(void)
 		.sub_ghz_link_config = app_get_sub_ghz_config(),
 	};
 
+	int err = bt_gatt_authorization_cb_register(&gatt_authorization_callbacks);
+	if (err) {
+		LOG_ERR("Registering GATT authorization callbacks failed (err %d)", err);
+		return;
+	}
```
ktaborowski, line 354: I think ble pal would be better than sample app, for calling Bluetooth api

totyz (approved):

PR #649, opened 2024-11-29
tests: refactor tests

ktaborowski:
[KRKNWK-14886]
Move and refactor tests:

- [x] critical_region
- [ ] crypto
- [x] crypto_keys
- [x] delay
- [x] interrupts
- [ ] ~~mfg~~ _not working_
- [ ] spi
- [ ] storage
- [x] temperature
- [x] time

## CI parameters

```yaml
Github_actions:
  NRF_revision: main

  Create_NRF_PR: false
Jenkins:
  test-sdk-sidewalk: master
```

## Description

JIRA ticket: 

## Self review

- [x] Change has been tested.

[KRKNWK-14886]: https://nordicsemi.atlassian.net/browse/KRKNWK-14886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests_new/integration/time/src/main.c
RobertGalatNordic, line 1: /**
 * split for 2 tests, it checks 2 interfaces, timer and uptime
 * 
 */

tests_new/integration/storage/boards/nrf54l15dk_nrf54l15_cpuapp_ns.overlay
RobertGalatNordic, line 1: should't be needed

tests_new/integration/storage/src/main.c
RobertGalatNordic, line 1: refactor tests, tests are coupled, need to be run in order, commented test

tests_new/integration/spi/src/main.c
RobertGalatNordic, line 1: /**
 * Test spi transfer, reuse external flash because it is already on DK? otherwise this test will require some fixture to run 
 * 
 */

tests_new/integration/mfg/src/main.c
RobertGalatNordic, line 1: /**
TODO rewrite the test

read suite:
v7
v8

write suite:

clear memory,
write values
read values
compare


persistant test
write values
dump flash to different address
initialize on new address (copy)
read values
compare

 */

tests_new/integration/crypto/src/main.c
RobertGalatNordic, line 1: #if defined(CONFIG_SOC_SERIES_NRF54LX) 
// TODO: this ifdef should be based on feature not SOC used. replace it by feature: SID_PAL_HASH_SHA512 check



 /**
I want to see what configurations were tested. Maybe some kind of matrix that show what function with what algorithms, and what values were tested
what error cases were covered

each function should have its own suite maybe some kind of macro generation of test functions would be usefull to make this information more compact

maybe separating functions, into separate files would be benefitial

maybe some defines, should be moved to Kconfig of the test, or separate configuration header.
Simmilarly there are many schemes like test_data, correct_response. extracting those long data arrays would make tests more redeable.
 */

tests_new/integration/crypto_keys/src/main.c
RobertGalatNordic, line 1: // keys_init repeat in every test, move to before_fn ? 

// deinit() repeat ( or should repeat) in all tests move to after_fn ?

tests_new/integration/delay/src/main.c
RobertGalatNordic, line 1: // TODO: split to 2 test cases, delay =0, delay > 0, add maybe cases with small delay, and big delay ... test for test for accuracy, print time stamps in fail message (to know why it failed)

RobertGalatNordic: make sure to use correct types, %lld for uint64 ( %d will get lower 32bits, so in values below uint32_max will print 0)

tests_new/integration/interrupts/src/main.c
RobertGalatNordic, line 1: add verification of the out buffer before starting swi, and after stop and deinit to check if swi was not triggered too many times, or after stop/deinit

tests_new/integration/mfg/testcase.yaml
RobertGalatNordic, line 1: Currently this test fails, and need to be rewritten.
Consider skipping this test, to allow this PR to merge, and plan rewriting it.

RobertGalatNordic (changes requested):
crypto
mfg
spi
storage

Require rewriting or major refactoring. Please move them to their old location, as they will be addressed in separate PRs.

If some test does not pass, please skip it and place it back in its old location (move to tests_new, only refactored/approved and working tests)

RobertGalatNordic (approved):

PR #653, opened 2024-12-04
cmake: removing manual linking of libgcc.a

tejlmand:
This commit removes the manual lookup and linking of libgcc.a.

Linking of C and runtime libraries are the responsibility of the higher level build system which handles the toolchain and building of the executable.

Current sidewalk implementation is broken in the sense that it fails to consider the CPU architecture and just looks up any libgcc.a file.

For example the code might discover `arm-zephyr-eabi/12.2.0/libgcc.a` when instead `arm-zephyr-eabi/12.2.0/thumb/v8-m.main+fp/hard/libgcc.a` should be used, and thus causiong link errors such as:
```
error: /.../12.2.0/libgcc.a(_arm_muldf3.o):
                                    conflicting CPU architectures 17/2
```
and:
```
error: zephyr/zephyr_pre0.elf uses VFP register arguments,
       /.../12.2.0/libgcc.a(_fixunsdfdi.o) does not
```

### DIFF
--- a/lib/cortex-m33/lora_fsk/CMakeLists.txt
+++ b/lib/cortex-m33/lora_fsk/CMakeLists.txt
@@ -6,11 +6,6 @@
 
 set(SID_LIB_DIR ${ZEPHYR_BASE}/../sidewalk/lib/cortex-m33/lora_fsk)
 
-set(GCC_LIB_DIR ${ZEPHYR_SDK_INSTALL_DIR}/arm-zephyr-eabi/lib/gcc/arm-zephyr-eabi)
-find_library(LIB_GCC NAMES gcc libgcc.a HINTS ${GCC_LIB_DIR}/* REQUIRED)
-
-zephyr_library_link_libraries(${SID_LIB_DIR}/libsid_protocol_all.a
-	${LIB_GCC}
-)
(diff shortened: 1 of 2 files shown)


nordicjm (approved):

carlescufi (approved):

ktaborowski:
Similar code for M 4 (nrf52) libraries: sidewalk/lib/cortex-m4/lora_fsk/CMakeLists.txt
Should it be also changes?

tejlmand:
> Similar code for M 4 (nrf52) libraries: sidewalk/lib/cortex-m4/lora_fsk/CMakeLists.txt Should it be also changes?

nice catch.
Sure it should, I didn't notice this was actually duplicate in both m33 and m4 when I was looking into nRF5340dk.

ktaborowski (approved):
lgtm

tejlmand:
> Similar code for M 4 (nrf52) libraries: sidewalk/lib/cortex-m4/lora_fsk/CMakeLists.txt Should it be also changes?

done

carlescufi:
@tejlmand @ktaborowski please apply this change to the `main` branch as well.

tejlmand:
> @tejlmand @ktaborowski please apply this change to the `main` branch as well.

there is already a PR for that here: https://github.com/nrfconnect/sdk-sidewalk/pull/654

PR #674, opened 2025-01-23
Build improvements

ktaborowski:
## CI parameters

```yaml
Github_actions:
  NRF_revision: main

  Create_NRF_PR: false
Jenkins:
  test-sdk-sidewalk: master
```

## Description

JIRA ticket: 

## Self review

- [ ] Change has been tested.


CMakeLists.txt
```diff
@@ -5,9 +5,11 @@
 #
 
 if(CONFIG_SIDEWALK_BUILD)
-	add_subdirectory(lib)
-	add_subdirectory(subsys)
-	add_subdirectory(utils)
+	if (CONFIG_SIDEWALK_USE_PREBUILTS)
```
RobertGalatNordic, line 8: nit: inconsistent indentation

Kconfig.dependencies
```diff
@@ -127,6 +180,18 @@ config PSA_WANT_ALG_CBC_PKCS7
 config PSA_WANT_ALG_HKDF
 	default n
 
+config PSA_USE_CC3XX_KEY_AGREEMENT_DRIVER
+	default n if SOC_NRF52840 || SOC_NRF5340_CPUAPP
```
RobertGalatNordic, line 184: I'm wondering if we could find a more generic reason for the if statement than specific SOC,
I thought that both nrf52 and nrf53 use crypto cell, so maybe this should be unconditionally default to false ? (as nrf54 do should not have and use this config? )

subsys/sal/common/sid_pal_types/sid_pal_common_ifc.h
RobertGalatNordic, line 1: Some tests fail because they do not see the headers that were moved.

RobertGalatNordic (approved):
Minor comments for future improvement.
Tests need to be fixed, but it seems it is a simple job of updating include path in tests.

Kconfig.dependencies
ktaborowski: the nrf crypto configuration is a minefield, I would rather not change this

PR #682, opened 2025-03-03
log: disable RTT backend when SHELL is in use.

RobertGalatNordic:
Logs are sent only to backends that are ready,
If some backend like SHELL takes some time to start, and other backend is already initialized.
The messages will not be sent to the uninitialled backend. The solution is to disable the unused backend
and therefore the logs will wait for the shell to initialize

## CI parameters

```yaml
Github_actions:
  NRF_revision: main

  Create_NRF_PR: false
Jenkins:
  test-sdk-sidewalk: master
  # or
  #  - filter1:
  #     board: nrf52
```

## Description

JIRA ticket: KRKNWK-20023

## Self review

- [x] Change has been tested.

### DIFF
--- a/samples/sid_end_device/Kconfig
+++ b/samples/sid_end_device/Kconfig
@@ -58,9 +58,6 @@ config SHELL_BACKEND_SERIAL_TX_RING_BUFFER_SIZE
 config SHELL_BACKEND_SERIAL_RX_RING_BUFFER_SIZE
 	default 1024
 
-config LOG_PROCESS_THREAD_STARTUP_DELAY_MS
-	default 100
-
 endif # SID_END_DEVICE_CLI
 
 config SID_END_DEVICE_AUTO_START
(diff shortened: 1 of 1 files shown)


samples/sid_end_device/Kconfig
```diff
@@ -173,4 +173,16 @@ config PSA_USE_CC3XX_KEY_MANAGEMENT_DRIVER
 
 endif #SIDEWALK_CRYPTO
 
+if SHELL
+# TODO: This is workaround for missing first few logs when SHELL backend is enabled together with other LOG backend.
+#       The logs in that case are not send to SHELL, as it is not ready yet, but pushed to other backends like RTT taht are avaliable.
+#       This change has been already introduced in mainline https://github.com/zephyrproject-rtos/zephyr/pull/85790/files
+#       Once this change is introduced in NCS, change in this file may be safely removed.
```
ktaborowski, line 180:
```suggestion
# TODO: Remove when merged in sdk-zephyr https://github.com/zephyrproject-rtos/zephyr/pull/85790/files 
# Workaround for missing first few logs.
```

ktaborowski (approved):
Shouldn't changes like this one be reverted?
https://github.com/nrfconnect/sdk-sidewalk/pull/679

samples/sid_end_device/Kconfig
RobertGalatNordic: good point about revert... I will add suggested comment change and revert the previous workaround.

BTW just an idea, this is not the first time when we introduce some temporary fix in our code that should be removed if some condition is met ( in that case change made to NCS copy of the zephyr) 
Maybe we could think about automating this process, maybe some auto notification, 
or regular CI run with reverted change and if all test pass ( assuming there was a test that could detect the issue) 
than it is safe to remove the change and PR should be created ... 
I can see a potential for automation this kind of workarounds

PR #750, opened 2026-05-14
sal: sid_pal: ble: keep disconnect cleanup on info errors

ktaborowski:
## Summary
- keep BLE disconnect teardown path resilient when `ble_conn_param_get()` fails during disconnect callback timing
- downgrade the connection-parameter fetch failure log in disconnect path and continue cleanup/notification flow
- add a unit test that covers the `bt_conn_get_info() -> -ENOTCONN` path and verifies disconnect callback plus connection cleanup

## Test plan
- [x] Run `tests/unit_tests/sid_ble_connection` locally
- [ ] Run DUT certification test `test_CERT_CB12[BLE]`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/subsys/sal/sid_pal/src/sid_ble_connection.c
+++ b/subsys/sal/sid_pal/src/sid_ble_connection.c
@@ -120,8 +120,7 @@ static void ble_disconnect_cb(struct bt_conn *conn, uint8_t reason)
 
 	int err = ble_conn_param_get(conn, &conn_params_prev);
 	if (err) {
-		LOG_ERR("Connection param get failed (err=%d)", err);
-		return;
+		LOG_WRN("Connection param get failed (err=%d)", err);
 	}
 
 	sid_ble_adapter_conn_disconnected((const uint8_t *)conn_data.addr);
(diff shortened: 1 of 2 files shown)


subsys/sal/sid_pal/src/sid_ble_connection.c
```diff
@@ -120,8 +120,7 @@ static void ble_disconnect_cb(struct bt_conn *conn, uint8_t reason)
 
 	int err = ble_conn_param_get(conn, &conn_params_prev);
 	if (err) {
-		LOG_ERR("Connection param get failed (err=%d)", err);
-		return;
+		LOG_WRN("Connection param get failed (err=%d)", err);
```
LuDuda, line 123: Does it only happen after moving to NCS v3.3.0

LuDuda (approved):

subsys/sal/sid_pal/src/sid_ble_connection.c
ktaborowski: No, but it's a relatively new code (added after add-on v1.1.0).

PR #752, opened 2026-05-18
fix(sid_pal/spi): use stable spi_dt_spec to avoid SPIM30 re-init bug

ktaborowski:
LR11xx WHOAMI / get_version returned 0x00 on nRF54L15 (NCS 3.3.0) while a direct spi_transceive_dt on the same bus worked. Root cause: zephyr nrfx SPIM uses pointer-equality on spi_config in spi_context_configured() to decide whether to skip reconfigure. The PAL passed its own static spi_config; the pointer mismatch forced nrfx_spim_uninit + nrfx_spim_init(skip_psel_cfg=true), which on SPIM30 leaves MISO stuck at 0x00.

Drop the hand-built spi_config and route every transfer through a single SPI_DT_SPEC_GET-derived spi_dt_spec so the driver stays on the fast path.

Also add an LR1110 GetVersion integration test that exercises the same bus end-to-end and would have caught this regression.

## CI parameters

```yaml
Jenkins:
  test-sdk-sidewalk: master
  # or
  #  - filter1:
  #     board: nrf52
```

## Description

JIRA ticket: 

## Self review

- [ ] Change has been tested.

### DIFF
--- a/subsys/sal/sid_pal/src/sid_pal_serial_bus_spi.c
+++ b/subsys/sal/sid_pal/src/sid_pal_serial_bus_spi.c
@@ -16,15 +16,26 @@
 #include <zephyr/logging/log.h>
 LOG_MODULE_REGISTER(sid_spi_bus, CONFIG_SPI_BUS_LOG_LEVEL);
 
-#define SPI_NODE DT_PARENT(DT_CHOSEN(zephyr_lora_transceiver))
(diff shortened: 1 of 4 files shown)


ktaborowski:
Continued in https://github.com/nrfconnect/sdk-sidewalk/pull/753

PR #753, opened 2026-05-19
Fix spi pal after migration to NCS v3.3.0

ktaborowski:
## CI parameters

```yaml
Jenkins:
  test-sdk-sidewalk: master
  # or
  #  - filter1:
  #     board: nrf52
```

## Description

Eliminating the hand-built `struct spi_config` so the pointer identity invariant the driver relies on is satisfied. +refactor.

`spi_transceive(ctx.device, &ctx.cfg, ...)` Replaced by `spi_transceive_dt(&bus_serial_spec, ...) ` -  idiomatic NCS/Zephyr pattern, which holds the same information, populated from device tree at compile time instead manually in source file.

Add test for lr1110 communication.

JIRA ticket: KRKNWK-21841

## Self review

- [x] Change has been tested.

### DIFF
--- a/subsys/sal/sid_pal/src/sid_pal_serial_bus_spi.c
+++ b/subsys/sal/sid_pal/src/sid_pal_serial_bus_spi.c
@@ -16,34 +16,24 @@
 #include <zephyr/logging/log.h>
 LOG_MODULE_REGISTER(sid_spi_bus, CONFIG_SPI_BUS_LOG_LEVEL);
 
-#define SPI_NODE DT_PARENT(DT_CHOSEN(zephyr_lora_transceiver))
+#define LORA_DT_NODE DT_CHOSEN(zephyr_lora_transceiver)
 
 #define SPI_OPTIONS                                                                                \
 	(uint16_t)(SPI_WORD_SET(8) | SPI_TRANSFER_MSB | SPI_OP_MODE_MASTER | SPI_FULL_DUPLEX)
 
-struct bus_serial_ctx_t {
-	const struct sid_pal_serial_bus_iface *iface;
-	const struct device *device;
-	struct spi_config cfg;
-};
+static const struct spi_dt_spec bus_serial_spec = SPI_DT_SPEC_GET(LORA_DT_NODE, SPI_OPTIONS);
 
-static sid_error_t bus_serial_spi_xfer(const struct sid_pal_serial_bus_iface *iface,
+static sid_error_t zephyr_spi_bus_xfer(const struct sid_pal_serial_bus_iface *iface,
 				       const struct sid_pal_serial_bus_client *client, uint8_t *tx,
 				       uint8_t *rx, size_t xfer_size);
-static sid_error_t bus_serial_spi_destroy(const struct sid_pal_serial_bus_iface *iface);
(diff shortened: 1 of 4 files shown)


tests/integration/spi/src/main.c
```diff
@@ -17,7 +19,10 @@
 #include <sid_pal_serial_bus_ifc.h>
 #include <sid_pal_serial_bus_spi_config.h>
 
-#define NULL_STRUCT_INITIALIZER { 0 }
+#define NULL_STRUCT_INITIALIZER                                                                    \
```
LuDuda, line 22: nit: style

subsys/sal/sid_pal/src/sid_pal_serial_bus_spi.c
```diff
@@ -103,16 +93,12 @@ sid_error_t sid_pal_serial_bus_nordic_spi_create(const struct sid_pal_serial_bus
...
 
-	if (!device_is_ready(bus_serial_ctx.device)) {
+	if (!spi_is_ready_dt(&bus_serial_spec)) {
 		LOG_ERR("SPI device not ready");
 		return SID_ERROR_IO_ERROR;
 	}
 
-	*iface = &bus_ops;
-	bus_serial_ctx.cfg.frequency = DT_PROP_OR(DT_CHOSEN(zephyr_lora_transceiver),
-						  spi_max_frequency, SPI_FREQUENCY_DEFAULT);
```
LuDuda, line 113: In new code, the default frequency is not set. Could this be a problem if customer won't have and frequency defined in DTS?

LuDuda (approved):

tests/integration/spi/src/main.c
```diff
@@ -117,4 +122,139 @@ ZTEST(spi_bus, test_only_tx_spi)
...
+	/*
+	 * LR1110 needs ~185ms to boot after RESET release before BUSY goes low.
+	 * Wait for BUSY low then add a small settle margin.
+	 */
+	zassert_ok(lr11xx_wait_busy(LR11XX_BUSY_TIMEOUT_MS),
+		   "LR11xx BUSY stuck high — chip not powered or reset not released");
+	k_msleep(5);
+
+	/*
+	 * Use raw Zephyr SPI (spi_transceive_dt) to eliminate PAL abstraction.
```
Damian-Nordic: Not sure why we want to eliminate PAL abstraction. The test was supposed to test the PAL abstraction as well, no? :)

Damian-Nordic (approved):

subsys/sal/sid_pal/src/sid_pal_serial_bus_spi.c
ktaborowski: the `spi_max_frequency` is `required`, so it should fail at build/devicetree generation if customer won't have it defined in DTS
https://github.com/nrfconnect/sdk-zephyr/blob/main/dts/bindings/spi/spi-device.yaml#L13-L16

tests/integration/spi/src/main.c
ktaborowski: Good catch, I used this test as spi playground. Now it uses the right API level

PR #782, opened 2026-07-13
sal: cmake: split common headers into interface libraries

ktaborowski:
Replace flat SAL common include directories with per-module INTERFACE libraries for sid_ifc and sid_pal_ifc. Consolidate legacy config and HAL sources into app_utils, fold ACE into sal/common, and update samples, tests, and CI scripts to link the new CMake targets.

## CI parameters

```yaml
Jenkins:
  test-sdk-sidewalk: master
  # or
  #  - filter1:
  #     board: nrf54l
```

## Description

JIRA ticket: 

## Self review

- [ ] Change has been tested.


samples/lr11xx_firmware_update/include/lr1110_transceiver_0401.h
LuDuda, line 1: Is this correct, that the FW changes with the same version? (i don't see diff on Github).

I would not recommend to do formating on files that do not belong to Nordic.

samples/sid_end_device/CMakeLists.txt
```diff
@@ -62,3 +62,20 @@ if(CONFIG_SID_END_DEVICE_CLI)
         src/cli/location_shell_events.c
     )
 endif()
+
+if(CONFIG_SIDEWALK_USE_PREBUILTS)
```
LuDuda: nit: But the KConfig name seems unnatural for me. Alternative could be like in OpenThread to use: `CONFIG_SIDEWALK_SOURCES` and revert logic, or `CONFIG_SIDEWALK_USE_PREBUILT_LIBRARIES`.

LuDuda (approved):
Indeed there is a lot of line changed :)

Went through the structure and CMakes, and looks solid, thanks.

samples/lr11xx_firmware_update/include/lr1110_transceiver_0401.h
ktaborowski: Done, thanks for pointing that out :)

samples/sid_end_device/CMakeLists.txt
ktaborowski: Renamed to `CONFIG_SIDEWALK_USE_PREBUILT_LIBRARIES`
btw as Sidewalk code base is not public unlike OpenThread, this Kconfig is mostly for internal purposes.

Damian-Nordic (approved):

LuDuda (approved):
LGTM :+1:

PR #788, opened 2026-07-17
samples: sid_end_device: add nRF70 Wi-Fi location PAL

FalcO0O:
Add a Sidewalk Wi-Fi PAL (sid_pal_wifi_ifc.h) backed by an asynchronous Zephyr net_mgmt Wi-Fi scan on the nRF70, enabling Wi-Fi-based location alongside (or instead of) the LR1110 GNSS/Wi-Fi PAL.

- subsys/nrf70_location: new nrf70_wifi_scan (net_mgmt scan wrapper) and sid_pal_wifi_nrf70_impl (Wi-Fi PAL) backend, enabled via the new CONFIG_SIDEWALK_WIFI_LOCATION_NRF70 Kconfig option (with scan band/type/timeout sub-options), and select NET_MGMT/NET_MGMT_EVENT/ NET_MGMT_EVENT_INFO so struct net_mgmt_event_callback::info is actually available to the scan callback.
- subsys/semtech: skip the sx126x_location NOSUPPORT stub and the LR1110's own Wi-Fi PAL entry points when the nRF70 PAL is active, so it takes priority while LR1110 keeps providing GNSS.
- Kconfig: tie the whole nRF70 Wi-Fi stack (WIFI, NETWORKING, WIFI_NRF70, NRF70_SCAN_ONLY, FW patch storage, and the scan band/type defaults) to the presence of the nrf7002eb2 shield via `if SHIELD_NRF7002EB2`, so `--shield nrf7002eb2` alone is enough - no -DFILE_SUFFIX or sysbuild SB_CONFIG_WIFI_NRF70* flags needed.
- samples/sid_end_device: no per-board wifi variant file needed either
  - the DUT/CLI shell variant (needed to drive `location scan`) is already the existing, board-agnostic overlay-dut.conf, applied the usual way (-DOVERLAY_CONFIG="overlay-dut.conf"). sysbuild_wifi.conf and the nRF54LM20 wifi board conf are both removed as redundant. Also a board overlay fix for the LED pin the shield repurposes as its SPI-CS line.
- west.yml / .gitignore: pull in the nrf_wifi module and ignore local provisioning artifacts.

Verified building and linking end to end (west build) for:
  nrf54lm20dk/nrf54lm20a/cpuapp --shield nrf7002eb2
  nrf54l15dk/nrf54l15/cpuapp (default, unaffected by this change)

## CI parameters

```yaml
Jenkins:
  test-sdk-sidewalk: master
  # or
  #  - filter1:
  #     board: nrf54l
```

## Description

JIRA ticket: 

## Self review

- [ ] Change has been tested.


samples/sid_end_device/boards/nrf54l15dk_nrf54l15_cpuapp.overlay
```diff
@@ -7,7 +7,7 @@
 /{
 	aliases {
 		state-notifier-connected = &led0;
-		state-notifier-time-sync = &led1;
+		/* led1 is repurposed as SPI-CS by the nrf7002eb2 Wi-Fi shield */
```
LuDuda: I don't think this is correct. We should have the dedicated overlay, which changes it only when nRF7002eb2 is enabled.

subsys/semtech/lr1110_location/sid_pal_gnss_wifi_impl.c
```diff
@@ -230,10 +230,12 @@ sid_error_t sid_pal_gnss_process_event(uint8_t event_id)
     return sid_pal_common_process_event();
 }
 
+#if !defined(CONFIG_SIDEWALK_WIFI_LOCATION_NRF70)
```
LuDuda: I don't think that LR1110 PAL implementation should depend on NRF70. If this is really necessary due to the build system, let's use different defines and maybe compile it out => @ktaborowski, thoughts?

subsys/semtech/CMakeLists.txt
```diff
@@ -67,7 +67,14 @@ target_compile_definitions(smtc_lbm INTERFACE
 )
 
 add_subdirectory_ifdef(CONFIG_SIDEWALK_SUBGHZ_RADIO_SX126X sx126x)
-add_subdirectory_ifdef(CONFIG_SIDEWALK_SUBGHZ_RADIO_SX126X sx126x_location)
+
+# When the nRF70 Wi-Fi PAL is active it provides the GNSS/Wi-Fi PAL symbols
+# instead of the sx126x_location NOSUPPORT stub, so skip the stub to avoid
+# duplicate symbol definitions.
+if(CONFIG_SIDEWALK_SUBGHZ_RADIO_SX126X AND NOT CONFIG_SIDEWALK_WIFI_LOCATION_NRF70)
```
LuDuda: @ktaborowski any other generic Config we could use rather than referencing to NRF70 here?

Kconfig
```diff
@@ -233,6 +233,139 @@ config SIDEWALK_USE_PREBUILT_LIBRARIES
...
+
+	  When SIDEWALK_SUBGHZ_RADIO_SX126X is also selected this PAL replaces
+	  the sx126x_location NOSUPPORT stub, so the (lora_fsk) Sidewalk library
+	  actually invokes the Wi-Fi PAL.
+
+	  This PAL can also coexist with SIDEWALK_SUBGHZ_RADIO_LR1110 (for
+	  example the nRF Sidewalk EB together with an nRF7002 EB): whenever WIFI
+	  is enabled the nRF70 provides the Wi-Fi PAL (taking priority), while the
+	  LR1110 keeps providing the GNSS PAL and the sub-GHz radio. When WIFI is
+	  disabled the LR1110 provides the Wi-Fi PAL itself.
```
LuDuda: I would love to not have direct dependencies between LR1110/SX1262. Maybe we can introduce new set of configs which would compile in/out certain features.

Kconfig
```diff
@@ -233,6 +233,139 @@ config SIDEWALK_USE_PREBUILT_LIBRARIES
...
+
+if SIDEWALK_WIFI_LOCATION_NRF70
```
LuDuda: Can we have these defines as part o subsys and nrf70 support instead?

Kconfig
```diff
@@ -233,6 +233,139 @@ config SIDEWALK_USE_PREBUILT_LIBRARIES
...
+	default y
+
+config NET_L2_ETHERNET
+	default y
+
+config NET_OFFLOAD
+	default y
+
+config WIFI_NRF70
+	default y
```
LuDuda: Can we imply this configs when `SIDEWALK_WIFI_LOCATION_NRF70` is enabled?

Kconfig
```diff
@@ -233,6 +233,139 @@ config SIDEWALK_USE_PREBUILT_LIBRARIES
...
+
+# NRF70_SCAN_ONLY and NRF_WIFI_PATCHES_EXT_FLASH_DISABLED are choice members
+# (NRF70_OPER_MODES / NRF_WIFI_PATCHES_EXT_FLASH_SUPPORT), so the default has
+# to be set on the choice itself, not on the member config.
+choice NRF70_OPER_MODES
+	default NRF70_SCAN_ONLY
+endchoice
+
+choice NRF_WIFI_PATCHES_EXT_FLASH_SUPPORT
+	default NRF_WIFI_PATCHES_EXT_FLASH_DISABLED
```
LuDuda: Is this not a default?

LuDuda (commented):
Overall looks good. Initial review.

samples/sid_end_device/boards/nrf54l15dk_nrf54l15_cpuapp.overlay
Damian-Nordic: Right, the problem is that there's no application-specific shield overlays support in Zephyr so I guess there's no way to do it without requiring a manual setting `EXTRA_DTC_OVERLAY_FILE` : https://github.com/zephyrproject-rtos/zephyr/pull/90469.

Ideally, the shield shouldn't delete the led but just mark it as disabled but I guess it's too late to change it for sidewalk :)

ktaborowski: nRF7002eb2 is a [shield](https://nrfconnectdocs.nordicsemi.com/ncs/latest/zephyr/hardware/porting/shields.html)
Please configure shield dts, not the board overlays

subsys/semtech/lr1110_location/sid_pal_gnss_wifi_impl.c
ktaborowski: Yes, this file implements [wifi pal interface](https://github.com/nrfconnect/sdk-sidewalk/blob/main/subsys/sal/common/public/sid_pal_ifc/wifi/sid_pal_wifi_ifc.h) for lr1110.
For nRF70xxx I recommend to create a new pal implementation, and a config to switch between them (e.g. based on shield, chosen in sample build?)

subsys/semtech/CMakeLists.txt
ktaborowski: I recommend to not edit this file at all - new wifi chip - new pal implementation file

ktaborowski (dismissed):
Needs work - separate lr1110 and nrf70 implementations, recommended to use shield to combine dts, configs and code together in one logical group, and avoid writing so many comments - use commit messages for changes justifications insted

LuDuda (commented):
Thanks for the iteration on splitting the PALs. I want to step back and align on the overall architecture, because I'd like us to avoid any coupling between the LR1110 and nRF70 ports, and ideally modifications in LR1110 location files.

Conceptually, LR1110 and nRF70 should not depend on each other at all. From Kconfig we should simply be able to select the location source:

- nRF70 location (Wi-Fi and GNSS stub), or
- LR1110 location (GNSS and Wi-Fi)

(We could also have generic hw-independent Wi-Fi and GNSS stubs)

With that model, nRF70 can be paired with either SX1262 or even LR1110 as the sub-GHz radio only, without pulling in the LR1110's own Wi-Fi/GNSS location path.

The nRF70 PAL should stand on its own and never require edits to the LR1110 implementation (no `#if !defined(CONFIG_SIDEWALK_WIFI_LOCATION_NRF70)` guards inside the LR1110 files, no changes to subsys/semtech/...).

The one genuinely tricky combination is nRF7002 for Wi-Fi + LR1110 for GNSS only. I see a few options there:

- Leave that combination for the customer to port themselves
- Introduce two new Kconfig options on the LR1110 side to independently enable/disable its Wi-Fi and GNSS features, so the Wi-Fi part can be compiled out cleanly while GNSS stay (as you sort of did)
- Cut off the sid_pal Wi-Fi or GNSS functions with some CMake magic (e.g. selectively excluding/overriding the relevant PAL objects at link time), instead of guarding the source with #ifs.

Option 2 is the nicer long-term answers, but if we go either route, that split must be contributed to the Sidewalk SDK upstream first - otherwise we're carrying a local fork of the LR1110 port and it becomes a maintenance burden for us.

For this PR, my preference is: keep the nRF70 location PAL fully self-contained (new files only, its own Kconfig, shield-driven DTS/config as @ktaborowski suggested) and don't touch the LR1110 port at all.

Does it make sense, or am i missing something?

CMakeLists.txt
```diff
@@ -4,6 +4,34 @@
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
 
+# Sidewalk-specific nrf7002eb2 overlays live under boards/shields/nrf7002eb2/.
+# Do not map state-notifier-time-sync to &led1 in board overlays; add it here only 
+# when the nrf7002eb2 shield is not used, and drop the alias from the shield overlay
+# when it is.
+if(DEFINED BOARD AND BOARD MATCHES "^nrf54l15dk_nrf54l15_cpuapp(_ns)?$")
```
Damian-Nordic: As discussed f2f, this mixes two different levels of software, i.e. this applies a shield-specific overlay to each sidewalk application although the overlay patches devicetree attributes used in a specific sample only (sid_end_device).
Unfortunately, Zephyr build system doesn't automatically pick an overlay up from `<sample-dir>/shields/<shield_name.overlay>` like it does for board overlays, so perhaps a CMake snippet like below needs to be added, but let it be part of the sample.

samples/sid_end_device/boards/nrf54l15dk_nrf54l15_cpuapp.overlay
Damian-Nordic: Note that the shield is defined in Zephyr so it's not easy to change it here.

subsys/nrf70_location/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+#
+# Copyright (c) 2026 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+zephyr_library_named(sid_pal_wifi_nrf70_impl)
```
Damian-Nordic: I think using `zephyr_library` is undesired as links into the firmware with `whole-archive` instead of just pulling necessary dependencies. It should be a regular CMake library and added as a dependency of `sid_sdk_component_pal_ifc` (can you confirm @ktaborowski?)

subsys/nrf70_location/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+if(NOT CONFIG_SIDEWALK_SUBGHZ_RADIO_LR1110)
+	target_sources(sid_pal_wifi_nrf70_impl PRIVATE
+		sid_pal_gnss_stub.c
+	)
+	target_link_libraries(sid_pal_wifi_nrf70_impl PRIVATE
+		sid_pal_gnss_ifc
+	)
+endif()
+
+target_include_directories(sid_pal_wifi_nrf70_impl PRIVATE
```
Damian-Nordic: This looks redundant. Headers are always looked up in the current dir as well.

subsys/nrf70_location/Kconfig
```diff
@@ -0,0 +1,65 @@
...
+config SIDEWALK_WIFI_LOCATION_NRF70_SCAN_TIMEOUT_MS
+	int "nRF70 Wi-Fi scan timeout (ms)"
+	default 20000
+	range 1000 60000
+	help
+	  Backstop timeout for a single nRF70 Wi-Fi scan.
+
+choice SIDEWALK_WIFI_LOCATION_NRF70_SCAN_BAND_FILTER
+	prompt "nRF70 Wi-Fi location scan band filter"
+	default SIDEWALK_WIFI_LOCATION_NRF70_SCAN_BAND_2_4GHZ if SHIELD_NRF7002EB2
```
Damian-Nordic: Aren't there any Kconfig available to check whether 5G is supported on the given chip (like nRF 7001) and select it accordingly? At least we need to make sure that all 7000/7001/7002 build with correct capabilities.

subsys/nrf70_location/Kconfig
```diff
@@ -0,0 +1,65 @@
...
+	bool "2.4 and 5 GHz"
+
+config SIDEWALK_WIFI_LOCATION_NRF70_SCAN_BAND_ALL
+	bool "All regulatory bands (2.4/5/6 GHz)"
+
+endchoice
+
+choice SIDEWALK_WIFI_LOCATION_NRF70_SCAN_TYPE
+	prompt "nRF70 Wi-Fi location scan type"
+	default SIDEWALK_WIFI_LOCATION_NRF70_SCAN_PASSIVE if SHIELD_NRF7002EB2
```
Damian-Nordic: I think we should stick to active if that works better even if we needed a different configuration for the power consumption report.

Damian-Nordic (commented):
Initial comments

subsys/nrf70_location/nrf70_wifi_scan.c
```diff
@@ -0,0 +1,197 @@
...
+	size_t stored;
+	size_t total;
+	int status;
+} scan_ctx;
+
+static atomic_t scan_done = ATOMIC_INIT(0);
+
+static struct net_if *wifi_iface(void)
+{
+	struct net_if *iface = net_if_get_first_wifi();
```
Damian-Nordic: `net_if_get_first_wifi` should be enough. If there's no wi-fi interface, there's no point in pretending the default is Wi-Fi.

subsys/nrf70_location/nrf70_wifi_scan.c
```diff
@@ -0,0 +1,197 @@
...
+	struct nrf70_wifi_scan_ap *r = scan_ctx.results;
+	size_t max = scan_ctx.max_results;
+	size_t pos;
+
+	if (!r || max == 0) {
+		return;
+	}
+
+	if (scan_ctx.stored < max) {
+		pos = scan_ctx.stored++;
```
Damian-Nordic, line 86: Isn't some deduplication needed? Don't you get the same AP multiple times in your tests?

subsys/nrf70_location/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+#
+
+zephyr_library_named(sid_pal_wifi_nrf70_impl)
+
+target_sources(sid_pal_wifi_nrf70_impl PRIVATE
+	nrf70_wifi_scan.c
+	sid_pal_wifi_nrf70_impl.c
+)
+
+if(NOT CONFIG_SIDEWALK_SUBGHZ_RADIO_LR1110)
```
Damian-Nordic: There should be Kconfig specific for nRF70 location to enable that to make it more future proof (in case we want to test it with other radios than lr1110).

subsys/nrf70_location/sid_pal_wifi_nrf70_impl.c
```diff
@@ -0,0 +1,230 @@
...
+	size_t stored = 0;
+	size_t total = 0;
+	int status = 0;
+
+	int ret = nrf70_wifi_scan_get_result(&stored, &total, &status);
+
+	if (ret == 0) {
+		/* Woken but scan not done yet: only act if the backstop timeout
+		 * has elapsed, otherwise wait for the real completion wake. */
+		if (k_uptime_get() >= scan_deadline) {
```
Damian-Nordic: if the backstop timeout hasn't elapsed, don't we want to request another event so that we ensure the abort is called on time? Is such a situation even possible?

subsys/nrf70_location/sid_pal_wifi_nrf70_impl.c
```diff
@@ -0,0 +1,230 @@
...
+
+	return SID_ERROR_NONE;
+}
+
+sid_error_t sid_pal_wifi_cancel_scan(void)
+{
+	/* Best-effort cancel. A scan that has not started yet (a deferred scan
+	 * still waiting for its delay to elapse) is cancelled cleanly. A scan
+	 * already running in the driver cannot be stopped in hardware (Zephyr
+	 * has no scan-cancel request); we just detach the listener so its
```
Damian-Nordic: The problem with this approach is that if we cannot cancel the request in the lower layers, then another scan request will fail anyway until the lower layer finishes. Even if it doesn't fail (because, say, another request is a deferred one), you would need to make sure that callback for the previous request is not interpreted as the callback for the current one.
Thus I wonder whether we couldn't just return `NOSUPPORT` to be more honest about our capabilities :).
Also, this and the presence of CONFIG_WIFI_NRF70_SCAN_TIMEOUT_S timeout in the lower layer, makes me wonder whether the backstop timer will not do more harm than good.

subsys/nrf70_location/nrf70_wifi_scan.h
```diff
@@ -0,0 +1,86 @@
...
+#endif
+
+/**
+ * A single access point discovered by a net_mgmt Wi-Fi scan.
+ *
+ * Only the fields the Sidewalk Wi-Fi PAL payload actually carries (BSSID and
+ * signal strength) are kept; SSID/channel/security are intentionally dropped to
+ * keep the collection buffer small.
+ */
+struct nrf70_wifi_scan_ap {
```
Damian-Nordic, line 21: Is there a reason this layer doesn't use `sid_pal_wifi_ap` directly to avoid translation between the two? The `mac_len` seems not to be usedfor anything important anyway.

Damian-Nordic (commented):
Looks good! Added a few more comments to make the solution a bit less complex and future-proof.

samples/sid_end_device/sample.yaml
```diff
@@ -157,6 +157,18 @@ tests:
       - cli
       - lr1110
 
+  sample.sidewalk.dut.nrf7002eb2:
+    platform_exclude:
+      - nrf54l15dk/nrf54l15/cpuapp/ns
+    extra_args:
+      - OVERLAY_CONFIG="overlay-dut.conf"
+      - SHIELD=nrf7002eb2
```
Damian-Nordic: I suggest that we test with `nrf7002eb2_nrf7000` by default as this is the best match for this use case.

samples/sid_end_device/sample.yaml
```diff
@@ -157,6 +157,18 @@ tests:
       - cli
       - lr1110
 
+  sample.sidewalk.dut.nrf7002eb2:
```
Damian-Nordic:
```suggestion
  sample.sidewalk.dut.ble_only.nrf7002eb2:
```

subsys/nrf70_location/Kconfig
```diff
@@ -0,0 +1,93 @@
...
+	  Backstop timeout for a single nRF70 Wi-Fi scan: only used if the
+	  net_mgmt scan-done event is never delivered. Keep this comfortably
+	  above (CONFIG_WIFI_NRF70_SCAN_TIMEOUT_S * 1000): the nRF70 driver
+	  has its own scan timeout, and this backstop firing first would
+	  report a failed scan to Sidewalk while the driver is still
+	  finishing the previous one, which then also delays the earliest
+	  possible next scan.
+
+# nRF7002 is dual-band (2.4/5 GHz); nRF7000/nRF7001 are 2.4 GHz only. There is
+# currently no Kconfig capability symbol distinguishing the variants, so the
```
Damian-Nordic: Is this true? I can see e.g. `SHIELD_NRF7002EB2_NRF7000`  in Zephyr

Damian-Nordic: Also, there seems to already be CONFIG_NRF70_2_4G_ONLY

Damian-Nordic (commented):
A few remaining comments regarding the configuration but it looks much better now :)

subsys/nrf70_location/nrf70_wifi_scan.h
FalcO0O: The driver reports RSSI as int8_t while sid_pal_wifi_ap::rssi is uint8_t, and this layer sorts/dedups by RSSI — sorting the unsigned field breaks for any value >= 0, which the nRF7x scan handler can produce.

subsys/nrf70_location/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+	)
+endif()
+
+# This PAL replaces the sx126x_location NOSUPPORT stub, which would otherwise
+# provide duplicate Wi-Fi symbols.
+if(TARGET sid_pal_radio_sx126x_location_impl)
+	set_target_properties(sid_pal_radio_sx126x_location_impl PROPERTIES
+		SOURCES ""
+		LINKER_LANGUAGE C
+	)
```
LuDuda: I propose to remove this and remove sx126x stub and have only:

- weaks for wifi and GNSS platform API location
- remove https://github.com/FalcO0O/sdk-sidewalk/blob/014ec21c91aece65b82a49f17c7b5f525277a82f/subsys/semtech/CMakeLists.txt#L70C1-L71C1
- `CONFIG_SIDEWALK_NRF7X_LOCATION` that enables `sid_pal_wifi_nrf70_impl.c`
- Introduce a new KConfig `CONFIG_SIDEWALK_LR1110_LOCATION` that enables https://github.com/FalcO0O/sdk-sidewalk/blob/014ec21c91aece65b82a49f17c7b5f525277a82f/subsys/semtech/CMakeLists.txt#L73C60-L73C75

subsys/nrf70_location/Kconfig
```diff
@@ -0,0 +1,53 @@
+#
+# Copyright (c) 2026 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+config SIDEWALK_WIFI_LOCATION_NRF70
+	bool "nRF70 Wi-Fi location PAL"
+	default y if SHIELD_NRF7002EB2 || SHIELD_NRF7002EB2_NRF7001 || SHIELD_NRF7002EB2_NRF7000 || SHIELD_NRF7002EB2_COEX
+	depends on !SIDEWALK_SUBGHZ_RADIO_LR1110
```
LuDuda: Then also remove this. Alternatively have depends on !CONFIG_SIDEWALK_LR1110_LOCATION for now.

The way to untangled dependency to LR1110_LOCATION probably would be to have some select option defined somewhere else? Not sure @Damian-Nordic.

LuDuda (commented):
Looks great :+1:

FalcO0O:
I tried the weak approach first. It does not work here, and the reason is that
the two sides sit on opposite sides of Zephyr's link line.

`zephyr_library_named()` archives are force-included: Zephyr wraps all of them
in `--whole-archive` (`zephyr/cmake/linker/ld/target.cmake`). The nRF70 PAL is a
plain `add_library`, so it lands after `--no-whole-archive` and is scanned
lazily, only for symbols that are still undefined. A weak definition in the stub
is therefore unconditionally part of the link, it satisfies the `sid_pal_wifi_*`
references coming from the prebuilt `libsidewalk_sdk.a`, and
`libsid_pal_wifi_nrf70_impl.a` is never opened at all - those references are the
only thing that would pull it in. A weak attribute only arbitrates between
definitions that are already in the link; it has no influence on whether an
archive gets opened. The build succeeds and Wi-Fi returns `NOSUPPORT`.

The `--start-group` in `cmake/sidewalk.cmake` does not change that either. It
comes from `target_link_options`, so it does not stay attached to the libraries
it was meant to enclose, and on the actual link line the group ends up holding a
single unrelated archive:

```
lib/cortex-m33/lora_fsk/libsidewalk_sdk.a                    <- before the group
-Wl,--start-group
  modules/sidewalk/subsys/semtech/sx126x/libsid_pal_radio_sx126x_impl.a
-Wl,--end-group
modules/sidewalk/subsys/nrf70_location/libsid_pal_wifi_nrf70_impl.a   <- after
```

I then tried `--whole-archive` to force the archive with the real implementation
into the link. That fails for the same class of reason: in
`target_link_libraries`, targets become nodes in the dependency graph and are
ordered and de-duplicated, while `"-Wl,..."` is an unrecognised string copied
through verbatim at the position where it was written.
`sid_pal_wifi_nrf70_impl` was also reachable through app -> `sidewalk` ->
`sidewalk_pal`, so CMake collapsed the occurrences and emitted the target ahead
of the flags, leaving an empty `--whole-archive`/`--no-whole-archive` pair. The
stub kept winning. Forcing the stub *out* this way was never an option either:
as a `zephyr_library` it sits inside Zephyr's own `--whole-archive` block no
matter what we add.

`$<LINK_LIBRARY:WHOLE_ARCHIVE,tgt>` does bind the flag to the target correctly,
but it needs CMake 3.24, above the 3.20 floor Zephyr declares.

However the outcome your proposal aims for is very neat, so to get that
behaviour I moved the decision from link time to compile time. Two files, each
built only when nothing implements the area it covers:

```cmake
if(NOT CONFIG_SIDEWALK_NRF7X_LOCATION AND NOT CONFIG_SIDEWALK_LR1110_LOCATION)   # sid_location_wifi.c
if(NOT CONFIG_SIDEWALK_LR1110_LOCATION)                                          # sid_location_gnss.c
```

Wi-Fi and GNSS need separate translation units because their conditions differ:
the nRF70 implements Wi-Fi but not GNSS. With a single combined file there is no
good option - compiling it contributes `sid_pal_wifi_*` and, by the mechanism
above, silently shadows the nRF70 implementation, while skipping it leaves
`sid_pal_gnss_*` unresolved. Source selection is the only granularity available
here, since both the archive and the object file are taken as a whole.

With exactly one definition of each symbol compiled in any configuration there
is nothing left for the linker to arbitrate, so the definitions are not marked
weak. On nrf54l15dk with the nRF70 shield the image now holds:

```
000667f1 T sid_pal_wifi_deinit     <- nrf70_location
0007bf3b T sid_pal_gnss_deinit     <- NOSUPPORT
```

The collision that required `set_target_properties(... SOURCES "")` is gone too.

subsys/nrf70_location/nrf70_wifi_scan.c
```diff
@@ -0,0 +1,220 @@
...
+{
+	/* Zephyr's Wi-Fi mgmt API has no scan-cancel request, so a scan already
+	 * running in the driver cannot be stopped; only stop listening to it. */
+	if (cb_registered) {
+		net_mgmt_del_event_callback(&scan_cb);
+		cb_registered = false;
+	}
+	scan_ctx.results = NULL;
+	scan_ctx.max_results = 0;
+	atomic_set(&scan_done, 0);
```
Damian-Nordic, line 219: nit: Not sure if this atomic is actually needed. These APIs are not thread-safe anyway and assumes there's one caller of these APIs. But ok, probably this is to address a misuse where nrf70_wifi_scan_get_result is called without waiting for the scan completion callback so no harm to keep it.

subsys/nrf70_location/sid_pal_wifi_nrf70_impl.c
```diff
@@ -0,0 +1,194 @@
...
+
+sid_error_t sid_pal_wifi_process_event(uint8_t event_id)
+{
+	ARG_UNUSED(event_id);
+
+	if (!scan_busy) {
+		return SID_ERROR_NONE;
+	}
+
+	if (!scan_started) {
```
Damian-Nordic, line 117: I would verify that `event_id == SID_PAL_WIFI_INTERNAL` in this case and `SID_PAL_WIFI_SCAN_COMPLETE` otherwise to make sure nothing goes off rails if there's deinit->init sequence while the underlying layer still processes the previous scan request or so.

subsys/sal/sid_pal/src/CMakeLists.txt
```diff
@@ -173,7 +173,7 @@ if(NOT CONFIG_SIDEWALK_PAL_ZEPHYR_LIBS_DISABLED AND CONFIG_SIDEWALK_DELAY)
 	target_link_libraries(sid_pal_delay_impl PRIVATE sid_pal_delay_ifc)
 endif()
 
-if(NOT CONFIG_SIDEWALK_PAL_ZEPHYR_LIBS_DISABLED AND CONFIG_SIDEWALK_SPI_BUS)
+if(NOT CONFIG_SIDEWALK_PAL_ZEPHYR_LIBS_DISABLED AND CONFIG_SIDEWALK_SPI_BUS AND NOT DEFINED ENV{SIDEWALK_TEST_SX126X_STUB})
```
Damian-Nordic: Testing leftover?

Kconfig
```diff
@@ -39,6 +39,13 @@ config SIDEWALK_SUBGHZ_RADIO_LR1110
 
 endchoice # SIDEWALK_SUBGHZ_RADIO
 
+config SIDEWALK_LR1110_LOCATION
```
Damian-Nordic:
```suggestion
config SIDEWALK_SUBGHZ_RADIO_LR1110_LOCATION
```
I think that LR1110 locationing depends on LR1110 as a whole (so worth naming it to indicate that).

Kconfig
```diff
@@ -39,6 +39,13 @@ config SIDEWALK_SUBGHZ_RADIO_LR1110
...
+	bool
+	default y if SIDEWALK_SUBGHZ_RADIO_LR1110
```
Damian-Nordic:
```suggestion
	default y
	depends on  SIDEWALK_SUBGHZ_RADIO_LR1110_LOCATION
```

Damian-Nordic (approved):
Looks neat! A few final remarks.

Damian-Nordic:
Also, you need to exclude nRF52 in sample.yaml

samples/sid_end_device/sysbuild.cmake
```diff
@@ -14,3 +14,14 @@ if(SB_CONFIG_BOARD_NRF54L15DK_NRF54L15_CPUAPP_NS)
...
+# The nrf7002eb2 shield repurposes the led1 pin as its SPI-CS line and deletes
+# the node on nRF54L15, so the state-notifier alias pointing at it has to go.
+if(SB_CONFIG_BOARD_NRF54L15DK_NRF54L15_CPUAPP OR SB_CONFIG_BOARD_NRF54L15DK_NRF54L15_CPUAPP_NS)
+  foreach(image sid_end_device mcuboot)
+    if("${SHIELD};${${image}_SHIELD}" MATCHES "nrf7002eb2")
+      sysbuild_cache_set(VAR ${image}_EXTRA_DTC_OVERLAY_FILE APPEND
+                         ${CMAKE_CURRENT_LIST_DIR}/shields/nrf7002eb2_nrf54l15dk.overlay)
+    endif()
+  endforeach()
+endif()
```
LuDuda, line 27: In general, i see that it follows a pattern about (which btw. don't know why is needed).

I think alternative is to write in the documentation that EXTRA_DTC_OVERLAY_FILE has to be appended. We try not to manually modify the SYSBUILD variables as whenever we used it before, we had to rewrite it :) 

But.. can live with it :)

LuDuda (approved):
Looks great! One thing missing is documentation, which we might do as a follow-up task.

subsys/sal/sid_pal/src/CMakeLists.txt
FalcO0O: yes :)

PR #790, opened 2026-07-21
ci: sync NCS toolchain in GitHub Actions and Docker

LuDuda:
## Summary

CI runs in a Docker image that ships with a fixed NCS toolchain. When west.yml bumps NCS (e.g. v3.3.0 → v3.4.0), jobs still pull the old :main image but fetch new source code - builds fail because the toolchain no longer matches.

**On PR:** a setup-ncs-toolchain step installs the toolchain required by the PR’s west.yml at runtime if the image is stale. PR CI stays green without rebuilding Docker first.

**After merge to main:** the Docker rebuild workflow runs when west.yml (or Docker files) change, so :main is updated with the matching toolchain for faster CI on later PRs.

**Older NCS on a PR:** if main already has NCS 3.4.0 and a PR still uses 3.3.0, Docker is not rebuilt on that PR. CI still uses :main and installs the 3.3.0 toolchain at runtime if needed. Docker rebuilds only when such a change is merged to main. The registry may keep multiple toolchain tags (e.g. :911f4c5c26, :fbf7391cab); CI always pulls :main.

## CI parameters

```yaml
Jenkins:
  test-sdk-sidewalk: master
  # or
  #  - filter1:
  #     board: nrf54l
```

## Description

JIRA ticket: 

## Self review

- [x] Change has been tested.

### DIFF
--- a/.github/actions/setup-ncs-toolchain/action.yml
+++ b/.github/actions/setup-ncs-toolchain/action.yml
@@ -1,0 +1,35 @@
+name: Setup NCS toolchain
+description: >
+  Make the NCS toolchain matching west.yml active for bash steps via BASH_ENV.
+  Call once after checkout (uses the pre-baked toolchain so west is available)
+  and once after 'west update' (installs the toolchain the manifest requires).
+
+runs:
+  using: composite
+  steps:
+    - name: Setup NCS toolchain
+      shell: bash
+      run: |
+        set -euo pipefail
+
+        # sdk-manager is installed in the image; BASH_ENV would hide it.
+        export NRFUTIL_HOME=/usr/local/share/nrfutil
+        NCS_INSTALL_DIR=/root/ncs
+
+        nrfutil install sdk-manager
+
+        if [ -x nrf/scripts/print_toolchain_checksum.sh ]; then
+          BUNDLE_ID=$(nrf/scripts/print_toolchain_checksum.sh)
+          if [ ! -d "${NCS_INSTALL_DIR}/toolchains/${BUNDLE_ID}" ]; then
+            nrfutil sdk-manager toolchain install \
+              --toolchain-bundle-id "${BUNDLE_ID}" --install-dir "${NCS_INSTALL_DIR}"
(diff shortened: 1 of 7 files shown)


.github/actions/setup-ncs-toolchain/action.yml
```diff
@@ -0,0 +1,35 @@
...
+        set -euo pipefail
+
+        # sdk-manager is installed in the image; BASH_ENV would hide it.
+        export NRFUTIL_HOME=/usr/local/share/nrfutil
+        NCS_INSTALL_DIR=/root/ncs
+
+        nrfutil install sdk-manager
+
+        if [ -x nrf/scripts/print_toolchain_checksum.sh ]; then
+          BUNDLE_ID=$(nrf/scripts/print_toolchain_checksum.sh)
```
ktaborowski, line 22: nit: use direct path `/workdir/nrf/scripts/print_toolchain_checksum.sh` for maintenability ?

ktaborowski (approved):

.github/docker/Dockerfile
```diff
@@ -7,21 +7,23 @@ ARG tools_version_minor=24
 ARG tools_version_patch=2
 
 ARG NRF_VERSION=main
+ARG TOOLCHAIN_BUNDLE_ID=
```
ktaborowski: nit: no need to `=` if no default value

LuDuda: @ktaborowski fixed.

.github/actions/setup-ncs-toolchain/action.yml
LuDuda: Thanks for the suggestion. I'd prefer to keep `nrf/scripts/print_toolchain_checksum.sh` here rather than switching to `/workdir/nrf/scripts/print_toolchain_checksum.sh`.

This action runs twice per job. Before west update and after. The `/workdir/nrf` is the image-baked copy and can be stale on NCS-bump PRs.

PR #798, opened 2026-07-30
sal: sid_pal: add DEV_ID_REG for nRF54LM20B

ktaborowski:
Builds for nrf54lm20dk/nrf54lm20b/cpuapp hit the "Unknown Device ID register." #error because the guard listed only NRF54LM20A_XXAA. nRF54LM20B has an identical FICR layout, with DEVICEID[2] at offset 0x04 of INFO, so it uses the same register.

## CI parameters

```yaml
Jenkins:
  test-sdk-sidewalk: master
  # or
  #  - filter1:
  #     board: nrf54l
```

## Description

JIRA ticket: 

## Self review

- [ ] Change has been tested.

### DIFF
--- a/subsys/sal/sid_pal/src/sid_mfg_storage.c
+++ b/subsys/sal/sid_pal/src/sid_mfg_storage.c
@@ -42,8 +42,7 @@ static int sid_mfg_storage_secure_read(uint16_t *p_value, uint8_t *buffer, uint1
 #ifndef DEV_ID_REG
 #if defined(NRF52840_XXAA) || defined(NRF52833_XXAA) || defined(NRF52832_XXAA)
 #define DEV_ID_REG (uint32_t)(NRF_FICR->DEVICEID[0])
-#elif defined(NRF54L15_XXAA) || defined(NRF54L10_XXAA) || defined(NRF54LV10A_XXAA) ||                \
-	defined(NRF54LM20A_XXAA)
+#elif defined(CONFIG_SOC_SERIES_NRF54L)
 #define DEV_ID_REG (uint32_t)(NRF_FICR->INFO.DEVICEID[0])
 #else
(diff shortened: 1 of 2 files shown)


Damian-Nordic (approved):

subsys/sal/sid_pal/src/sid_mfg_storage.c
```diff
@@ -40,10 +40,8 @@ static int sid_mfg_storage_secure_read(uint16_t *p_value, uint8_t *buffer, uint1
 #define FLASH_MEM_CHUNK (128)
 
 #ifndef DEV_ID_REG
-#if defined(NRF52840_XXAA) || defined(NRF52833_XXAA) || defined(NRF52832_XXAA)
-#define DEV_ID_REG (uint32_t)(NRF_FICR->DEVICEID[0])
-#elif defined(NRF54L15_XXAA) || defined(NRF54L10_XXAA) || defined(NRF54LV10A_XXAA) ||                \
-	defined(NRF54LM20A_XXAA)
+#if defined(NRF54L15_XXAA) || defined(NRF54L10_XXAA) || defined(NRF54LV10A_XXAA) ||                \
+	defined(NRF54LM20A_XXAA) || defined(NRF54LM20B_XXAA)
```
LuDuda: Looks good. I wonder, if we can use NRF54L_FAMILY define instead.

`#if defined(CONFIG_SOC_SERIES_NRF54L)`

LuDuda (approved):

subsys/sal/sid_pal/src/sid_mfg_storage.c
LuDuda: Ahh.. and you need rebase to get nRF52840 support back.

PR #802, opened 2026-08-17
scripts: add a west command to provision device

Damian-Nordic:
Add "west sid provision" command that is a wrapper on top of provision.py script that automatically detects the mfg_storage partition address and flashes the resulting hex onto the board.

## CI parameters

```yaml
Jenkins:
  test-sdk-sidewalk: master
  # or
  #  - filter1:
  #     board: nrf54l
```

## Description

JIRA ticket: 

## Self review

- [ ] Change has been tested.

### DIFF
--- a/scripts/west_commands/sid.py
+++ b/scripts/west_commands/sid.py
@@ -1,0 +1,262 @@
+#!/usr/bin/env python3
+# Copyright (c) 2026 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+
+from __future__ import annotations
+
+import json
+import re
+import subprocess
+import sys
+import tempfile
+import yaml
+
+from pathlib import Path
+from west.commands import WestCommand
+
+
+def resolve_app_build_dir(build_dir: Path) -> Path:
+    if (build_dir / "zephyr" / "zephyr.dts").is_file():
+        return build_dir
+
+    domains_file = build_dir / "domains.yaml"
+    if not domains_file.is_file():
+        raise FileNotFoundError(f"no domains.yaml found under {build_dir}")
+
+    data = yaml.safe_load(domains_file.read_text())
+    default = data["default"]
+    domain_dirs = {d["name"]: Path(d["build_dir"]) for d in data["domains"]}
+    if default not in domain_dirs:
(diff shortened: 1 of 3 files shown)


scripts/west_commands/sid.py
```diff
@@ -0,0 +1,262 @@
...
+
+    proc = subprocess.run(cmd, check=False)
+    if proc.returncode:
+        sys.exit(proc.returncode)
+
+
+def program_device(hex_path: Path, serial_number: str) -> None:
+    proc = subprocess.run(
+        [
+            "nrfutil",
```
ktaborowski, line 140: No handling for missing `nrfutil`?

scripts/west_commands/sid.py
```diff
@@ -0,0 +1,262 @@
...
+    print(f"Multiple devices connected ({len(devices)}):")
+    for idx, serial in enumerate(devices, 1):
+        print(f"  {idx}. {serial}")
+
+    prompt = f"Select device (1-{len(devices)}): "
+    while True:
+        try:
+            choice = int(input(prompt))
+        except (EOFError, ValueError):
+            continue
```
ktaborowski, line 97: possible looping forever here? maybe try once more, and if still fails, ends script with error?

ktaborowski (approved):
looks good

scripts/west_commands/sid.py
Damian-Nordic: Yes, if a user always types an incorrect value :). This the same behavior as in `west flash`.

scripts/west_commands/sid.py
Damian-Nordic: You will get an error, but right, we can instruct a user where to find the tool etc.

ktaborowski: It's a minor, non -blocking. If the error logs are descriptive enough, no need to change

scripts/west_commands/sid.py
ktaborowski: I missed it's user input, all good now

scripts/west_commands/sid.py
Damian-Nordic: Ok, so I'll merge this as-is and we can make it more informative in the next iteration.

PR #805, opened 2026-08-26
doc: update documentation before v1.3.0 release

Damian-Nordic:
Update release notes, migration guide, compatibility matrix and other doc pages where neceesary.

## CI parameters

```yaml
Jenkins:
  test-sdk-sidewalk: master
  # or
  #  - filter1:
  #     board: nrf54l
```

## Description

JIRA ticket: KRKNWK-22369

## Self review

- [ ] Change has been tested.


doc/releases_and_migration/release_notes_addon_v130.rst
```diff
@@ -25,6 +25,10 @@ Changelog
        On the nRF54L Series platforms, enabling this feature required disabling write protection for the ``mfg_storage`` partition, due to the limited number of configurable protected regions.
        If you lock the bootloader with an alternative method, such as the ``UICR.BOOTCONF`` register, set the Kconfig option ``CONFIG_FPROTECT`` to ``y`` to restore write protection for the ``mfg_storage`` partition.
 
+  * :ref:`nRF Sidewalk EB <nrf_sidewalk_eb>` shield support for LoRa and FSK on the nRF54L15 DK and nRF54LM20 DK.
+  * :ref:`lr11xx_firmware_update` sample for updating Semtech LR1110 transceiver firmware.
+  * :ref:`application_development` documentation section.
+
```
ktaborowski, line 33: minor: west sid provison command added (missing doc in setup section?)

doc/releases_and_migration/migration_guide_addon_v130.rst
```diff
@@ -1,13 +1,54 @@
...
    :local:
    :depth: 2
 
-.. _migration_changelog_mcuboot_rsa:
+This guide assists you in migrating from Amazon Sidewalk Add-On v1.2.0 to v1.3.0.
+
+Overview
+********
+
+Amazon Sidewalk Add-On v1.3.0 upgrades the nRF Connect SDK to v3.4.0, aligns MCUboot with NCS security recommendations, and moves persistent Sidewalk key storage to KMU on nRF54L Series platforms.
```
annwoj: Small nit. In links there's one link that still redirects to 3.3.0. I believe it should be 3.4.0 then? https://github.com/nrfconnect/sdk-sidewalk/blob/main/doc/links.rst#L33

doc/releases_and_migration/release_notes_addon_v130.rst
```diff
@@ -25,6 +25,10 @@ Changelog
        On the nRF54L Series platforms, enabling this feature required disabling write protection for the ``mfg_storage`` partition, due to the limited number of configurable protected regions.
```
annwoj: I'm not sure if I'm understanding this correctly, but I think this reads a bit ambiguous. 
I read it as: 
- ``enabling this feature (CONFIG_FPROTECT) requires disabling write protection for mfg_storage``
but also
- ``set CONFIG_FPROTECT to y to restore write protection for mfg_storage``
which seems a bit contradictory if read that way? 
Or maybe it's a completely missed understanding on my end.
But the other understanding that I could pull from it would be:

```suggestion
       On the nRF54L Series devices, this feature consumes one of the configurable protected regions, which are limited in number.
       For this reason, write protection for the ``mfg_storage`` partition is disabled by default when the feature is enabled.
       If you make the bootloader immutable by another mechanism, such as the ``UICR.BOOTCONF`` register, you can disable MCUboot's ``CONFIG_FPROTECT`` and set ``CONFIG_FPROTECT`` to ``y`` in the application image to restore write protection for the ``mfg_storage`` partition.
```

doc/releases_and_migration/release_notes_addon_v130.rst
```diff
@@ -44,3 +48,14 @@ Changelog
 * Fixed:
 
   * KRKNWK-20863: Increased boot time due to bootloader configuration.
+  * KRKNWK-21160: Semtech interrupt line stuck high during Wi-Fi location scan.
+  * KRKNWK-22208: Bluetooth LE advertising failure after Sidewalk deinit and reinit cycle.
+
```
ktaborowski, line 58: Bugfix Sidewalk libraries fixes link pause/resume during GNSS scans to resolve Semtech LR11xx hardware resource conflicts, fixes issues like:
* KRKNWK-20851: Cannot initialize location when LoRa or FSK transport is initialized in Amazon Sidewalk
* KRKNWK-20856: GNSS scan enters endless loop
* KRKNWK-21159: Crash in sid_location_run due to invalid context in Wi-Fi callback

ktaborowski (approved):

doc/releases_and_migration/release_notes_addon_v130.rst
```diff
@@ -25,6 +25,10 @@ Changelog
        On the nRF54L Series platforms, enabling this feature required disabling write protection for the ``mfg_storage`` partition, due to the limited number of configurable protected regions.
        If you lock the bootloader with an alternative method, such as the ``UICR.BOOTCONF`` register, set the Kconfig option ``CONFIG_FPROTECT`` to ``y`` to restore write protection for the ``mfg_storage`` partition.
```
annwoj:
```suggestion
```

doc/releases_and_migration/release_notes_addon_v130.rst
```diff
@@ -25,6 +25,10 @@ Changelog
...
 
+  * :ref:`nRF Sidewalk EB <nrf_sidewalk_eb>` shield support for LoRa and FSK on the nRF54L15 DK and nRF54LM20 DK.
```
annwoj:
```suggestion
  * :ref:`nRF Sidewalk EB <nrf_sidewalk_eb>` shield support for the nRF54L15 DK and nRF54LM20 DK, in both LoRa and FSK modes.
```

doc/releases_and_migration/release_notes_addon_v130.rst
```diff
@@ -44,3 +48,14 @@ Changelog
 * Fixed:
 
   * KRKNWK-20863: Increased boot time due to bootloader configuration.
+  * KRKNWK-21160: Semtech interrupt line stuck high during Wi-Fi location scan.
+  * KRKNWK-22208: Bluetooth LE advertising failure after Sidewalk deinit and reinit cycle.
```
annwoj: would align it to the format of other points
```suggestion
  * Long boot time of the device caused by the bootloader configuration. (KRKNWK-20863)
  * Semtech interrupt line remaining in the high state during a Wi-Fi location scan. (KRKNWK-21160)
  * Bluetooth LE advertising failing to start after a Sidewalk deinitialization and reinitialization cycle. (KRKNWK-22208)
```

doc/releases_and_migration/release_notes_addon_v130.rst
```diff
@@ -44,3 +48,14 @@ Changelog
...
+
+Backward compatibility
+**********************
+
+This release preserves backward compatibility for the application-facing Sidewalk API.
+For migration details, see the :ref:`migration_guide_addon_v130` page.
+
+.. note::
+   While this release mostly maintains API compatibility, ensure to test your applications and board configurations thoroughly after upgrading.
```
annwoj: a bit awkward placement, plus maybe not necessary
```suggestion
   While this release mostly maintains API compatibility, ensure to test your applications and board configurations after upgrading.
```

doc/releases_and_migration/migration_guide_addon_v130.rst
```diff
@@ -1,13 +1,54 @@
...
+
+Key updates in this release include:
+
+* Updating the nRF Connect SDK from v3.3.0 to v3.4.0.
+* MCUboot downgrade protection, runtime flash locking, and NCS-default signature types (see :ref:`migration_130_mcuboot_rsa`).
+* Persistent Sidewalk key storage in KMU on nRF54L Series platforms (see :ref:`migration_130_kmu`).
+* Removal of nRF53 Series platform support (see :ref:`migration_130_removed_platforms`).
```
annwoj: I wonder if this is necessary. Maybe it should just point to release notes instead? User might expect these changes will be explained on this page, while it's only the 2nd bullet that has some migration information, namely the part about signatures.

doc/releases_and_migration/migration_guide_addon_v130.rst
annwoj: Shouldn't we point to specific version?

annwoj, line 78: So I guess that means Keys don't have to be reprovisioned? Should this be mentioned clearly?

annwoj: Since I don't think these are consecutive steps
```suggestion
When encountering issues, check the following:

* For linker error or boot issues, perform a pristine build:
```

annwoj:
```suggestion
* For other issues, ensure all dependencies are up to date:
```

doc/releases_and_migration/migration_guide_addon_v130.rst
```diff
@@ -32,6 +73,8 @@ Perform a pristine build of the application after adding this override.
 .. note::
    For new designs, use the NCS default signature type: ED25519 on nRF54L Series platforms, ECDSA with P-256 curve on nRF52 Series platforms.
```
annwoj: Maybe this could be included somehow  in line 57? It already mentions this

annwoj (changes requested):

doc/releases_and_migration/migration_guide_addon_v130.rst
Damian-Nordic: This follows the pattern used in previous release, but I agree. Will replace with a link to release notes.

doc/releases_and_migration/migration_guide_addon_v130.rst
Damian-Nordic: Right, so I'll remove this.

doc/releases_and_migration/migration_guide_addon_v130.rst
Damian-Nordic: CONFIG_SIDEWALK_CRYPTO_PSA_KEY_STORAGE_KMU=n just disables the new feature so it keeps working like in the old days. This line just explains why changing this setting on devices in the field won't work.

Damian-Nordic: I used defensive language here because even though currently it seems that the last supporting version is 1.2.0, we can't be sure there will never be any 1.2.1, so this doc would get outdated in such a case. Does it make sense or should I add that it's 1.2.0 for the time being?

doc/releases_and_migration/release_notes_addon_v130.rst
Damian-Nordic: I don't think this is contradictory, because the second sentence `set CONFIG_FPROTECT to y to restore write protection for mfg_storage` is under a condition, so the formal structure of this fragment is:
1. If you keep using FPROTECT in bootloader, you can't use FPROTECT in the application (this is our default)
2. If you don't use FPROTECT in bootloader (because you use another method), we advise that you restore FPROTECT in the application.

Perhaps, I can add a clarification that on nRF54L, you can't use FPROTECT in both bootloader and the application.

doc/releases_and_migration/release_notes_addon_v130.rst
Damian-Nordic: I omitted that because I didn't add it to the documentation yet so I assumed it's not worth mentioning undocumented feature.

doc/releases_and_migration/release_notes_addon_v130.rst
Damian-Nordic: The shield kinda adds the support for LoRa and FSK modes, so how about I split it into two lines:
```
:ref:`nRF Sidewalk EB <nrf_sidewalk_eb>` shield support for the nRF54L15 DK and nRF54LM20 DK.
The shield adds the support for LoRa and FSK link types.
```
?

doc/releases_and_migration/release_notes_addon_v130.rst
Damian-Nordic: Yeah, this is copied from previous releases but sounds a bit like truism to me so I will drop this line.

doc/releases_and_migration/release_notes_addon_v130.rst
annwoj: Thanks! let's add the clarification then, I think it could be useful

doc/releases_and_migration/release_notes_addon_v130.rst
annwoj: We can split it  - would it still be correct if we said:

``Support for the :ref:`nRF Sidewalk EB <nrf_sidewalk_eb>` shield on the nRF54L15 DK and nRF54LM20 DK.
The shield enables the LoRa and FSK link types.`` 
?
Just to avoid repetitions

doc/releases_and_migration/migration_guide_addon_v130.rst
annwoj: Yeah, I know it does. In this specific case though I think it showed that the solution isn't ideal, so we can update from now on

doc/releases_and_migration/migration_guide_addon_v130.rst
annwoj: Than maybe we could say something along the lines: The version earlier than v1.3.0?

annwoj: got it, thanks!

annwoj (approved):

PR #825, opened 2026-09-18
Fix on-device certification cannot store the manufacturing page on nRF54L

ktaborowski:
## CI parameters

```yaml
Jenkins:
  test-sdk-sidewalk: master
  # or
  #  - filter1:
  #     board: nrf54l
```

## Description

JIRA ticket: KRKNWK-22489

## Self review

- [x] Change has been tested.


subsys/sal/sid_pal/include/sid_mfg_storage.h
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * @brief Flush the staged manufacturing data.
+ *
+ * Writes from sid_pal_mfg_store_write() are staged in RAM and become durable
+ * only when this function is called. Staged data is discarded by
+ * sid_pal_mfg_store_erase() and sid_pal_mfg_store_deinit(), and is kept when
+ * the flush fails, so that it can be retried.
+ *
+ * @return 0 on success, -ERRNO on error, or SID_ERROR_NOSUPPORT when built
```
Damian-Nordic, line 20: Could you say why we actually only support writing TLVs when SIDEWALK_MFG_STORAGE_DIAGNOSTIC is enabled? The Kconfig is not well documented so it's hard to determine what it meant to do.

subsys/sal/sid_pal/src/sid_mfg_storage.c
```diff
@@ -55,6 +58,71 @@ static const struct device *flash_dev;
...
+		return 0;
+	}
+
+	if (!flash_dev || tlv_flash.end_offset <= tlv_flash.start_offset) {
+		return -EINVAL;
+	}
+
+	const uint32_t size = mfg_store_size();
+
+	mfg_stage = sid_hal_malloc(size);
```
Damian-Nordic, line 91: Just a hint that an alternative could be using https://docs.zephyrproject.org/latest/doxygen/html/group__stream__flash.html that would avoid using a large buffer to store all and use a small write buffer instead. But I'll leave it to you to decide if it's worth changing.

Damian-Nordic (approved):

subsys/sal/sid_pal/include/sid_mfg_storage.h
ktaborowski: Good point, will rename the config

doc/known_issues.rst
```diff
@@ -26,6 +26,17 @@ A known issue can list one or more of the following entries:
...
 *******************************
 
+KRKNWK-22489: On-device certification cannot store the manufacturing page
+
+  The ``cert store`` command of the on-device certification CLI (``CONFIG_SIDEWALK_ON_DEV_CERT``) returns ``SID_ERROR_NOSUPPORT`` (-6) and the generated credentials are not stored.
+  The manufacturing storage writes each TLV entry with a 4-byte alignment, while the RRAM write block size is 16 bytes, so the flash driver rejects the write.
+
+  **Failing Test Cases:**
+
+  - None.
```
annwoj, line 36: If none, do we need to mention it?
```suggestion
```

ktaborowski: It's my intuition that it's better to explicitly state, there is no failing test case (so this issue doesn't impact qualification).

annwoj: Fair enough, if you think there's a value added to this information, let's leave as is

annwoj (approved):

Damian-Nordic (approved):

subsys/sal/sid_pal/src/sid_mfg_storage.c
ktaborowski: Good idea, but requires refactor (re-implement) sid_od_dev_cert module. Stays with ram buffer to apply the fix

